PR #1, opened 2016-02-24
Add SecPolicy suitable for validating TLS.

Lukasa:
I'm looking at writing a tool that allows validation of certificate chains using the appropriate binary available on the system. As a first pass, I want to try this on OS X, which involves making some function calls and types that aren't yet exposed in this library.

The first of those is the SecPolicy and SecPolicyRef. In this case, for my limited use-case, all I need is the `SecPolicyCreateSSL()` call, so that's all I've exposed here: essentially an opaque struct that can be created but then has almost nothing done with it. Admittedly quite boring, but turns out to be all we need.

Coming down the line I may add one or two other methods, but otherwise this is pretty complete.

Let me know if you need anything extra from me as part of this work!

### DIFF
--- a/security-framework/src/policy.rs
+++ b/security-framework/src/policy.rs
@@ -9,6 +9,7 @@ use std::fmt;
 
 use ErrorNew;
 use base::{Error, Result};
+use secure_transport::ProtocolSide;
 
 make_wrapper! {
     /// A type representing a certificate validation policy.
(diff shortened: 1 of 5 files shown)


security-framework/src/policy.rs
```diff
@@ -0,0 +1,52 @@
...
+impl fmt::Debug for SecPolicy {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("SecPolicy")
+           .finish()
+    }
+}
+
+impl SecPolicy {
+    /// Creates a `SecPolicy` suitable for validating certificates for SSL.
+    pub fn for_ssl(server: bool, hostname: &str) -> Result<SecPolicy> {
```
sfackler: I try to avoid using `bool` for parameters like this since I can never remember what `SecPolicy::for_ssl(true, "foo.com")` means vs `SecPolicy::for_ssl(false, "foo.com")`. The SecureTransport wrappers use a `ProtocolSide` enum that we could probably use here as well. https://github.com/sfackler/rust-security-framework/blob/master/security-framework/src/secure_transport.rs#L104

sfackler:
Looks good to me other than the one comment, thanks!

security-framework/src/policy.rs
Lukasa: Sounds good. =) Change made.

Lukasa:
Ok, feedback applied. =)

sfackler:
Thanks!

PR #2, opened 2016-02-25
Add SecTrustCreateWithCertificates binding.

Lukasa:
Follows on from #1.

This adds a binding to the stunningly convenient SecTrustCreateWithCertificates function, which can be used in combination with the policy added in #1 to validate TLS certificate chains.


Lukasa:
If you're interested in what I'm building with this, check out [rust-certitude](https://github.com/Lukasa/rust-certitude).

sfackler:
Looks good, thanks!

One thing I've wanted to do is to make a crate that wraps SChannel, SecureTransport, and OpenSSL on the relevant platforms to give you a TLS implementation that "just works" without having to worry about installing OpenSSL headers or whatever. It looks like rust-certitude should be a great thing to use there.

PR #4, opened 2016-03-09
Add `Send` bound to SecIdentity

### LINKED ISSUE #3: `Sync` and/or `Send` not implemented for various types
I'm getting the following error in a project of mine:

```
src/main.rs:132:9: 132:22 note: `*mut security_framework_sys::base::OpaqueSecIdentityRef` cannot be shared between threads safely
src/main.rs:132:9: 132:22 note: required because it appears within the type `security_framework::identity::SecIdentity`
src/main.rs:132:9: 132:22 note: required because it appears within the type `Config`
```

I'm trying to fetch the identity from the keychain once, and then save it / pass it around to various threads that need it.  If `Security.framework` supports it, could you add the appropriate bounds?

andrew-d:
Fixes #3

### DIFF
--- a/security-framework/src/identity.rs
+++ b/security-framework/src/identity.rs
@@ -51,3 +51,22 @@ impl SecIdentity {
         }
     }
 }
+
+unsafe impl Send for SecIdentity {}
+
+
+#[cfg(test)]
+mod test {
+    use std::ptr;
+
+    use super::SecIdentity;
+
+    #[test]
+    fn identity_has_send_bound() {
+        // We use a pointer so we can get a null pointer to the unconstructable SecIdentity type.
+        fn assert_send<T: Send>(_: *const T) {}
+
+        let ptr: *const SecIdentity = ptr::null();
+        assert_send(ptr);
+    }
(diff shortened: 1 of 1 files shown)


security-framework/src/identity.rs
```diff
@@ -51,3 +51,22 @@ impl SecIdentity {
...
+
+
+#[cfg(test)]
+mod test {
+    use std::ptr;
+
+    use super::SecIdentity;
+
+    #[test]
+    fn identity_has_send_bound() {
```
sfackler: If you cfg-off the test to OSX, you can get a `SecIdentity` here: https://github.com/sfackler/rust-security-framework/blob/master/security-framework/src/os/macos/mod.rs#L21, but an easier way is to do something like this:

```rust
fn assert_send<T: Send>() {}

assert_send::<SecIdentity>();
```

andrew-d: Oh, that's cool - switched to using that :-)

sfackler:
Thanks!

PR #5, opened 2016-04-17
Test on multiple versions of OS X.

frewsxcv:


sfackler:
If I had to guess, the failure with XCode 7.1 might be fixed by altering the test logic to maintain the SecKeychain while the SecIdentity retrieved from it is in use.

PR #8, opened 2016-06-07
Make Secure Transport error into an enum.

frewsxcv:


frewsxcv:
Was just experimenting what this would look like. @sfackler not sure if you have any opinions regarding this.

security-framework/src/secure_transport.rs
```diff
@@ -673,7 +787,7 @@ impl SslContext {
         unsafe {
             let ret = SSLSetIOFuncs(self.0, read_func::<S>, write_func::<S>);
             if ret != errSecSuccess {
-                return Err(HandshakeError::Failure(Error::new(ret)));
+                return Err(HandshakeError::Failure(ErrorCode::from_i32(ret).unwrap()));
```
frewsxcv: `unwrap` isn't great, but the only other alternative I can think of is to create a catch-all `UnknownError(i32)` variant on the `ErrorCode` enum.

security-framework/src/secure_transport.rs
```diff
@@ -99,6 +99,120 @@ use cipher_suite::CipherSuite;
 use identity::SecIdentity;
 use trust::{SecTrust, TrustResult};
 
+
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+#[repr(i32)]
+pub enum ErrorCode {
```
frewsxcv: This is a pretty crappy name.

frewsxcv:
I converted one method to use the new error type in 509f9e4d31737c403b25d62bc4182deac50c5b79. It could certainly be done in more, but I'm not going to spend any more time on it if this change is not desired.

sfackler:
I'm not sure this is the right way to go with respect to error reporting. The set of possible error codes is pretty hard to capture in a comprehensive manner, particularly as new OS versions come out - the unwrap in the conversion logic is an example of where that goes bad. I also don't expect that people will want to match over all possible errors vs checking a specific one or two.

What might make more sense would be to define consts in security-framework that just reference the consts in security-framework-sys but with more rustic names.

frewsxcv:
Sounds reasonable, thanks for the input.

PR #9, opened 2016-07-16
discard import-export from s-f on ios

kali:
import-export module is skipped in -sys crate, it should be discarded too from the "rusty" crate to build on ios.

As far as I can tell, without this change, current top of tree does not crosscompile to ios.

### DIFF
--- a/security-framework/src/lib.rs
+++ b/security-framework/src/lib.rs
@@ -63,6 +63,7 @@ pub mod base;
 pub mod certificate;
 pub mod cipher_suite;
 pub mod identity;
+#[cfg(target_os = "macos")]
 pub mod import_export;
 pub mod item;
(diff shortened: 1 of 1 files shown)


sfackler:
Oops! `SecImportExport.h` does appear in the iOS SDK, so I think we'll actually want to remove the `#[cfg(target_os = "macos")]` from the module in the sys crate.

sfackler:
Ah, but it doesn't include the stuff we're using.

sfackler:
This should un-break the build. Thanks!

sfackler:
It looks like we now ship iOS toolchains - I'll update the Travis configuration to at least build for iOS.

kali:
great news, thank a lot :)

PR #15, opened 2016-09-30
Allow calling functions that may not exist

alexcrichton:
This crate uses Cargo features to define the API, gating access to APIs
which weren't enabled. This isn't necessarily always the desired use
case, however. Sometimes a binary wants to be compiled as maximally
flexible (e.g runs almost anywhere) but still call newer APIs where it
can.

To accommodate this use case, this commit implements a fallback for APIs
where features are not enabled. Currently this is just a
proof-of-concept for one function to gauge interest, but the theory is
that this could generally apply to many APIs to provide a uniform API
for this crate across all OS versions.

Technically speaking, this enables a feature where when an FFI function
wouldn't otherwise be defined we define our own. Our own wrapper then
looks up the symbol at runtime (via `dlopen` + `dlsym`) and caches the
result in a global. The wrapper then also provides a fallback
implementation which typically just returns an error. This is similar to
what's done in the standard library as well.

### DIFF
--- a/security-framework/src/secure_transport/ffi.rs
+++ b/security-framework/src/secure_transport/ffi.rs
@@ -1,0 +1,59 @@
+#![allow(bad_style)]
+#![allow(unused_imports)]
+#![allow(dead_code)]
+
+use core_foundation_sys::base::*;
+use security_framework_sys::base::*;
+pub use security_framework_sys::secure_transport::*;
+
+use std::mem;
+use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT, Ordering};
+use libc;
+
+unsafe fn get<'a, T>(ptr: &'a AtomicUsize, name: &str) -> Option<&'a T> {
+    assert_eq!(mem::size_of::<T>(), mem::size_of_val(ptr));
+    match ptr.load(Ordering::SeqCst) {
+        0 => {}
+        1 => return None,
+        _ => return Some(&*(ptr as *const _ as *const T)),
+    }
+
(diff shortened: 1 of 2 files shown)


alexcrichton:
I initially thought of doing this [based on your comment](https://github.com/sfackler/rust-native-tls/pull/9#discussion_r81423831) that we should disable SSLv2/3 on platforms by default (which makes sense to me). The `set_protocol_version_min` method however is only defined on 10.8, so reasonable semantics to me would be that native-tls works on 10.7 but is just "less secure", but native-tls also works on 10.8+ but is as secure as it can be?

Does all that make sense?

sfackler:
So, to work around the specific issue for native-tls, we can use a separate function that I mentioned on that PR.

This does seem like an interesting idea, though, and is something that seems like it'd be useful in a broader variety of crates, or, at least rust-openssl as well.

alexcrichton:
Would you like me to expand this to APIs that otherwise require a feature to opt-in to?

sfackler:
Can we chat briefly about this after libs triage today? I think this could be a cool approach but just want to talk over the implications in a higher bandwidth context.

sfackler:
Or before triage for that matter.

alexcrichton:
Sure, I'll ping you beforehand

alexcrichton:
Ok, conclusion offline was to not add this just yet. The major downside is that you can silently rely on functionality on a newer OS without realizing it (e.g. entire APIs). 

For the specific use case in https://github.com/sfackler/rust-native-tls/pull/9 we can use older deprecated APIs.

PR #16, opened 2016-10-04
Bind `SSLSetProtocolVersionEnabled`

alexcrichton:
This may be useful if OSX 10.7 support is desired or otherwise working
with older OSX versions.


sfackler:
Thanks!

PR #17, opened 2016-10-16
Add support for generic password functions.

neonquill:
I'm not sure if this is the right place to make this change (vs. creating an entirely new library), so I am completely open to recommendations on how to make this more useful.

Also, I wasn't sure how to correctly call `CFRelease` on a `SecKeychainItemRef` variable. In the end, I removed the Opaque version in `security-framework-sys/src/base.rs`.  Better alternatives welcome.
- Able to get, set, modify, and delete generic passwords in the
  keychain.
- Only works with the default keychain search list.
- Currently no support for internet passwords.
- All tests pass.


security-framework-sys/src/base.rs
```diff
@@ -8,7 +8,10 @@ pub type SecKeychainRef = *mut OpaqueSecKeychainRef;
 
 #[repr(C)]
 pub struct OpaqueSecKeychainItemRef(c_void);
-pub type SecKeychainItemRef = *mut OpaqueSecKeychainItemRef;
+//pub type SecKeychainItemRef = *mut OpaqueSecKeychainItemRef;
+// XXX Need to do this to allow converting from a SecKeyChainItemRef
+// XXX to CFTypeRef.
+pub type SecKeychainItemRef = *mut c_void;
```
sfackler: You can simply cast between pointers of different types.

```rust
let x = 0 as SecKeychainItemRef;
let y = x as SecIdentityRef;
```

security-framework-sys/src/passwords.rs
```diff
@@ -0,0 +1,34 @@
+use core_foundation_sys::base::{OSStatus, CFTypeRef};
+use libc::{c_char, c_void};
+
+use base::{SecKeychainItemRef, SecKeychainRef};
+
+extern "C" {
+    #[cfg(target_os = "macos")]
+    pub fn SecKeychainFindGenericPassword(keychainkeychainOrArray: CFTypeRef,
```
sfackler: These should be placed in the module corresponding to the header in which they are defined - `keychain` in this case I believe.

security-framework/src/os/macos/passwords.rs
```diff
@@ -0,0 +1,166 @@
...
+use libc::c_void;
+
+use cvt;
+use base::Result;
+
+/// Find a generic password in the default list of keychains.
+///
+/// The underlying system supports passwords with 0 values, so this
+/// returns a vector of bytes rather than a string.
+pub fn find_generic_password(service: &str, account: &str) -> Result<Vec<u8>> {
```
sfackler: It seems like these should be made a bit more powerful to cover the functionality the underlying function provides. In particular, the optional specification of a keychain list and the option to retrieve the keychain item.

security-framework/src/os/macos/passwords.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    #[test]
+    fn missing_password() {
+        let password = find_generic_password("this_service_does_not_exist",
+                                             "this_account_is_bogus");
+        assert!(password.is_err());
+    }
+
+    #[test]
+    fn round_trip_password() {
```
sfackler: It'd probably be nice to have these use a temporary keychain to avoid polluting the system one.

neonquill:
Thanks for taking the time to give me feedback!

I think I've addressed most of your concerns.  The one thing I'm most unsure of is returning the `SecKeychainItem` from `find_generic_password.`  I'm not entirely sure returning a tuple is a good way to return the password and the item.  I'm also not very clear about the correct way to handle the reference counting (`wrap_under_create_rule` vs. `wrap_under_get_rule`).

neonquill:
I meant to also point out that I left in the tests which use the default keychain, rather than specifying a temporary one.  I completely understand my concern, but it felt like hiding those tests behind a feature flag was a reasonable compromise.  They don't run by default, but I'm still able to test the default use case for the functions.

steveatinfincia:
Hi :)

What is left to do on this PR, and can I help with any of it?

neonquill:
I would be very happy to have some help :)

As far as I know there are two high level issues.
1. There are a couple of potential issues (marked with XXX) in the code.  Most importantly, I wasn't able to figure out how the underlying reference counting worked (passwords.rs:65).  I suspect I just haven't found the right documentation explaining when to use get vs. create.
2. I suspect my changes need to be updated for the latest version of rust.

I'm happy to have somebody else take this over, or provide some guidance.  I don't have a lot of time to do the research, but I'd be happy to spend some time doing the grunt work.

steveatinfincia:
@neonquill Your branch still compiles here :)

I can probably clear up the get vs. create stuff and how it relates to the `wrap_under_get_rule` and `wrap_under_create_rule` calls, because they're a little confusing, but they do the right thing if you match the words with whatever you called (when that is possible, more below).

Short version: `passwords.rs:65` is using the correct wrap function, because we already own the object and don't want to retain it again and cause a leak.

Long version: in CoreFoundation, any time you call a function that includes the words "create" or "copy", the object you get back in the return value has a retain count of `1`, you own it and must release it later on. Any other function, or one with "get" in the name, you generally don't own the object that is returned, and must retain it yourself if you don't want it to disappear at the wrong time.

[`wrap_under_create_rule`](https://github.com/servo/core-foundation-rs/blob/44e872a427120f2e35c4d6ca4e02ee33f0ca798a/core-foundation/src/lib.rs#L37) doesn't try to retain the object again, because we should already own it and that would lead to a retain count of `2`, a memory leak.

[`wrap_under_get_rule`](https://github.com/servo/core-foundation-rs/blob/44e872a427120f2e35c4d6ca4e02ee33f0ca798a/core-foundation/src/lib.rs#L24) *will* retain the object for you so that it doesn't get freed while you're using it. If instead you call `wrap_under_create_rule` for something you don't own, you will end up with a use-after-free bug.

Now, for `SecKeychainFindGenericPassword()`, we can't really go by the function name (it doesn't actually *return* an object, which is likely why the create/get rule doesn't apply), however the [documentation](https://developer.apple.com/reference/security/1397301-seckeychainfindgenericpassword?language=objc) for it says that the `itemRef` parameter is the callers responsibility to release at some point, which means the retain count is already `1` so we should use `wrap_under_create_rule` there to avoid a leak.

neonquill:
@steveatinfincia  Wow!  Thanks for the CoreFoundation lesson, that all makes a lot of sense!

I've removed that comment from the code.  I will take some time this weekend to revisit the comments from @sfackler and see if there's anything else I can do to clean up the code.

neonquill:
I've gone through and attempted to clean everything up.  As far as I can tell, I've addressed all of the requested changes from @sfackler.  Sorry if I've missed anything.

In case it's preferable, I've rebased against the current master and squashed the commits: https://github.com/neonquill/rust-security-framework/commit/41df830fadfbca0b7c7bf340942aa31a39ede3c7

steveatinfincia:
I believe I found a use-after-free bug here, exposed by the `cross_keychain_corruption_temp` test, which was segfaulting every few times I ran it due to the array of keychains being released too soon. 

As far as I can tell, it's because the functions that take an array of keychains were creating a `CFArray` wrapper from them inside a match arm, but only returning the inner `CFTypeRef` from it, and then using it later on.

Since the `CFArray` itself isn't returned from the match arm, it gets dropped immediately when that scope ends. When the `CFArray` wrapper gets dropped, it calls `CFRelease()` on the inner `CFTypeRef`, which is then freed because nothing else has retained it.

Rust would normally warn about this, but `as_CFTypeRef()` [transmutes and returns](https://github.com/servo/core-foundation-rs/blob/master/core-foundation/src/lib.rs#L32) the inner `CFTypeRef`, so Rust doesn't know that it's a reference to the wrapper rather than an owned type.

I've pushed a fix [here](https://github.com/infincia/rust-security-framework/commit/c00a545656fc68bf1047f44fede6e7403e8d412c)

neonquill:
I've resolved the current conflicts with master and pulled in the fixes from @steveatinfincia.

If there's anything else I can do to move this forward, please let me know.  I'm also happy to close this PR and let @steveatinfincia take over with his version.

joshbooks:
howdy y'all, I'd love to have the functionality in this PR but I'm not sure where it got stalled, anything I can do to help?

steveatinfincia:
@sfackler when you have some free time, please let me know if there are any other changes you would need before this PR can be merged.

I would be happy to open a new one if that would be easier.

archaelus:
I recently used this functionality, and it's really useful for building other CLI API clients on OS X (takes care of credential storage pretty nicely). What would it take to get this merged, or the functionality pulled in in some other way?

jakeswenson:
Whats the status of this PR? i'd also like to use this feature to store passwords from a cli app

kornelski:
I've merged this feature based on somewhat modified version of this PR.

PR #19, opened 2016-11-29
Add support for TLS client certificates

andrew-d:
This allows using an identity in the keychain as a SSL/TLS client certificate.


sfackler:
Thanks!

PR #20, opened 2016-11-30
Add ability to search for items by label

adunham-stripe:
This is useful if we know the label of the thing we're searching for already.  I tested this with the following example:

```rust
extern crate security_framework;

use security_framework::item::{ItemSearchOptions, ItemClass};

pub fn main() {
    let results = ItemSearchOptions::new()
        .class(ItemClass::Identity)
        .label("adunham")
        .search()
        .unwrap();

    println!("results = {:?}", results);
}
```

This shows the expected output:

```
results = [SearchResult { reference: Some(Identity(SecIdentity { certificate: SecCertificate { subject: "adunham" }, private_key: SecKey })) }]
```

### DIFF
--- a/security-framework-sys/src/item.rs
+++ b/security-framework-sys/src/item.rs
@@ -20,6 +20,7 @@ extern "C" {
     pub static kSecMatchSearchList: CFStringRef;
 
     pub static kSecAttrKeyType: CFStringRef;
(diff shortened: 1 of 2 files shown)


andrew-d:
(aaaaand apparently I used my work account to create the PR instead - whoops!)

sfackler:
Thanks!

PR #21, opened 2016-12-08
SSLSetProtocolVersionEnabled is macos only

kali:
It may impact whatever @alexcrichton  in sfackler/rust-security-framework#16 was trying to achieve.


sfackler:
Oops, thanks!

PR #23, opened 2016-12-15
Add ClientBuilder::handshake2 for async clients

alexcrichton:
Currently the ClientBuilder::handshake function doesn't work with asynchronous
clients as it just does a loop internally. This function extracts that logic
into a new `MidHandshakeClientBuilder` type and `ClientHandshakeError` enum to
handle the loop externally.

This should allow asynchronous clients to still use `ClientBuilder` and the
`anchor_certificates` function, namely allowing `native-tls` to update to using
`ClientBuilder` and leveraging this logic.

### DIFF
--- a/security-framework/src/secure_transport.rs
+++ b/security-framework/src/secure_transport.rs
@@ -130,6 +130,27 @@ pub enum HandshakeError<S> {
     Interrupted(MidHandshakeSslStream<S>),
 }
 
+impl<S> From<Error> for HandshakeError<S> {
+    fn from(err: Error) -> HandshakeError<S> {
+        HandshakeError::Failure(err)
+    }
+}
+
+/// An error or intermediate state after a TLS handshake attempt.
+#[derive(Debug)]
+pub enum ClientHandshakeError<S> {
+    /// The handshake failed.
+    Failure(Error),
+    /// The handshake was interrupted midway through.
+    Interrupted(MidHandshakeClientBuilder<S>),
+}
+
+impl<S> From<Error> for ClientHandshakeError<S> {
+    fn from(err: Error) -> ClientHandshakeError<S> {
+        ClientHandshakeError::Failure(err)
+    }
+}
+
 /// An SSL stream midway through the handshake process.
 #[derive(Debug)]
 pub struct MidHandshakeSslStream<S> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Ah this also adds a few assorted other methods used by native-tls to just export the options of `SslContext` on `ClientBuilder`.

The motivation here is to update native-tls with an `anchor_certificates` function in one form or another!

security-framework/src/secure_transport.rs
```diff
@@ -192,6 +213,82 @@ impl<S> MidHandshakeSslStream<S> {
     }
 }
 
+/// An SSL stream midway through the handshake process.
+#[derive(Debug)]
+pub struct MidHandshakeClientBuilder<S> {
+    stream: Option<MidHandshakeSslStream<S>>,
```
sfackler: Why does this need to be an `Option`?

security-framework/src/secure_transport.rs
```diff
@@ -992,21 +1096,54 @@ impl ClientBuilder {
         self
     }
 
+    /// Use the specified chain of certificates for extra certificates in the
+    /// chain if a custom `identity` is also specified through the `identity`
+    /// method.
+    pub fn identity_chain(&mut self, chain: &[SecCertificate]) -> &mut Self {
```
sfackler: I would probably just call this `certificate_chain`.

alexcrichton:
Updated!

alexcrichton:
Hm ok I pushed another update that should fix the compilation on iOS, but it also raises the question of what to do about protocol configuration on iOS which doesn't have `set_protocol_version_enabled` but does have `set_protocol_version_{min,max}` presumably? Do you think we should infer the min/max?

sfackler:
Might make sense to switch to min/max everywhere and then conditionally compile the implementation.

security-framework/src/secure_transport.rs
```diff
@@ -992,21 +1102,47 @@ impl ClientBuilder {
...
+    pub fn protocol_max(&mut self, max: SslProtocol) -> &mut Self {
+        self.protocol_max = Some(max);
+        self
+    }
+
     /// Initiates a new SSL/TLS session over a stream connected to the specified
     /// domain.
+    ///
+    /// Note that this method assumes that the stream `S` is in *blocking* mode,
+    /// and it will spin unnecessarily if the stream is set to nonblocking mode.
```
sfackler: I think it'll just fail rather than spin?

alexcrichton: Good point! Updated

PR #25, opened 2017-02-28
make SslAuthenticate enum Copy+Clone

bluejekyll:
This is a simple enum, and will make functions that take this as a parameter easier to use.

Also, it looks like the same treatment could/should be done to SslClientCertificateState, but I haven't run into needing that.


sfackler:
Thanks!

PR #28, opened 2017-03-29
check buffered_read_size before calling SSLRead

seanmonstar:
cc https://github.com/seanmonstar/reqwest/issues/26

### DIFF
--- a/security-framework/src/secure_transport.rs
+++ b/security-framework/src/secure_transport.rs
@@ -83,6 +83,7 @@ use security_framework_sys::base::{errSecSuccess, errSecIO, errSecBadReq, errSec
                                    errSecNotTrusted};
 use security_framework_sys::secure_transport::*;
 use std::any::Any;
+use std::cmp;
 use std::io;
 use std::io::prelude::*;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


security-framework/src/secure_transport.rs
```diff
@@ -1292,7 +1296,7 @@ mod test {
         p!(ctx.set_peer_domain_name("google.com"));
         let stream = p!(TcpStream::connect("google.com:443"));
         let mut stream = p!(ctx.handshake(stream));
-        p!(stream.write_all(b"GET / HTTP/1.0\r\n\r\n"));
+        p!(stream.write_all(b"GET / HTTP/1.1\r\nHost: google.com\r\n\r\n"));
```
seanmonstar: By setting the version to 1.1, Google should keep the connection alive, and reading to the end (without the buffered check) should then hang. Cannot confirm test on my equipment.

seanmonstar:
Updated to not set len longer than buf.len.

seanmonstar:
https://github.com/sfackler/rust-security-framework/commit/4cd518d11a5c7e292a48c7ced15555e931bd0278 already landed, closing.

PR #29, opened 2017-04-26
Unconditionally define protocol versions

jrmuizel:
Not having these can cause panics in negotiated_protocol_version()
if run on a system newer than security-framework was configured
for.


sfackler:
Thanks! I filed #30 to fix the enumness in the next breaking release.

PR #31, opened 2017-04-26
Add a client example

jrmuizel:


sfackler:
Could this mirror the existing example in the rustdoc: http://sfackler.github.io/rust-security-framework/doc/v0.1.10/security_framework/secure_transport/index.html#examples?

In particular, I'd like to encourage use of `ClientBuilder` rather than `SslContext` directly since that can be pretty tricky to configure.

jrmuizel:
I'll take a look.

jrmuizel:
I switched to ClientBuilder

sfackler:
Thanks!

PR #32, opened 2017-04-27
Add some of the functions needed to get certificates

jrmuizel:


sfackler:
Thanks! Were you planning on piping them through to security-framework itself or is this good to go?

jrmuizel:
I was thinking of piping it through, but haven't gotten around to it. This at least makes it possible to get the information from the -sys version if I never get around to finishing it.

sfackler:
Cool, seems reasonable to land this as is either way.

PR #33, opened 2017-05-11
Add option to not trust built-in certs

jchien14:
Exposes a `trust_anchor_certificates_only` method on `ClientBuilder` that allows for specifying whether to trust the built-in root certificates or not. Defaults to trusting the root certificates for backwards compatibility. Allows for exclusive certificate pinning. 

Added a unit test that should throw an error as there are no available anchor certificates.


sfackler:
Thanks! Sorry for missing this.

PR #34, opened 2017-05-16
Implement get certificate count + get certificate at index

jchien14:
- Implements `get_certificate_count` and `get_certificate_at_index` on SecTrust to retrieve specific certificates in the chain. Unit tests added for both.
- Add the 10.12 feature, and bindings for `SecKeyCopyExternalRepresentation` and `SecItemExport`.


security-framework-sys/src/trust.rs
```diff
@@ -21,7 +21,7 @@ pub type SecTrustRef = *mut __SecTrust;
 extern "C" {
     pub fn SecTrustGetTypeID() -> CFTypeID;
     pub fn SecTrustGetCertificateCount(trust: SecTrustRef) -> CFIndex;
-    pub fn SecTrustGetCertificateAtIndex(trust: SecTrustRef) -> SecCertificateRef;
+    pub fn SecTrustGetCertificateAtIndex(trust: SecTrustRef, ix: CFIndex) -> SecCertificateRef;
```
sfackler: Ugh, I need to get ctest set up :(

security-framework/src/trust.rs
```diff
@@ -105,6 +106,29 @@ impl SecTrust {
             Ok(TrustResult::from_raw(result))
         }
     }
+
+    /// Returns the number of certificates in an evaluated certificate chain.
+    ///
+    /// Note: evaluate must first be called on the SecTrust.
+    pub fn get_certificate_count(&self) -> Result<i64> {
```
sfackler: Can you remove the `get_` prefixes here and below?

security-framework/src/trust.rs
```diff
@@ -105,6 +106,29 @@ impl SecTrust {
...
+    }
+
+    /// Returns a specific certificate from the certificate chain used to evaluate trust.
+    ///
+    /// Note: evaluate must first be called on the SecTrust.
+    pub fn certificate_at_index(&self, ix: i64) -> Result<SecCertificate> {
+        unsafe {
+            let certificate = SecTrustGetCertificateAtIndex(self.0, ix);
+            if certificate.is_null() {
+                Err(Error::from_code(errSecParam))
```
sfackler: Seems like it'd make more sense to return an option here?

security-framework/src/trust.rs
```diff
@@ -105,6 +106,29 @@ impl SecTrust {
             Ok(TrustResult::from_raw(result))
         }
     }
+
+    /// Returns the number of certificates in an evaluated certificate chain.
+    ///
+    /// Note: evaluate must first be called on the SecTrust.
+    pub fn certificate_count(&self) -> Result<i64> {
+        unsafe {
+            Ok(SecTrustGetCertificateCount(self.0))
```
sfackler: Why does this return an result if it's always `Ok`?

security-framework/src/trust.rs
```diff
@@ -105,6 +106,29 @@ impl SecTrust {
             Ok(TrustResult::from_raw(result))
         }
     }
+
+    /// Returns the number of certificates in an evaluated certificate chain.
+    ///
+    /// Note: evaluate must first be called on the SecTrust.
+    pub fn certificate_count(&self) -> i64 {
```
sfackler: `i64` isn't going to be the right type on 32 bit, right? It probably makes sense to just return CFIndex directly.

mako-taco:
Can we get this landed? I have to use a fork of this repo right now...

sfackler:
The i64 use for certificate indices needs to be adjusted I believe.

security-framework/src/trust.rs
cjubb39: `isize` seems like it'd fit here?.  Also have a fork of the repo -- would like to unify :)

jchien14:
Fixed! Although probably not a big deal for macOS since Rust only supports 10.7+ and 10.6 was the last supported 32-bit OS. Sorry for the delay

sfackler:
Thanks!

PR #35, opened 2017-05-24
Enable specification of certain cipher suites

jchien14:
This enables implementation of https://github.com/sfackler/rust-native-tls/issues/4 for Mac.

Instead of just a `supported_ciphers` call, I decided to enable both whitelist/blacklist so the API allows people to disable (i.e. blacklist) very insecure cipher suites without needing to list every other suite.

On the rust-native-tls layer, we'll maintain a mapping from the `Algorithm` to a Vector of CipherSuite that corresponds to that Algorithm.

### DIFF
--- a/security-framework/src/cipher_suite.rs
+++ b/security-framework/src/cipher_suite.rs
@@ -7,7 +7,7 @@ macro_rules! make_suites {
     ($($suite:ident),+) => {
         /// Specifies cipher suites
         #[allow(non_camel_case_types, missing_docs)]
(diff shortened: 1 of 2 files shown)


jchien14:
Hmm, my inclination to fix up the test so that it always passes even if no available ciphers are available to make the connection is to make a test only method that unconditionally returns the Stream even on handshake failure. This would allow testing of other stream configuration in the builder as well. 

Does that sound reasonable to you?

Issue #36, opened 2017-05-24
Pkcs12ImportOptions::import panics if optional items are missing from pkcs12 file


azdlowry:
If any of the items returned from SecPKCS12Import don't include all of kSecImportItemLabel, kSecImportItemKeyID, kSecImportItemTrust, kSecImportItemCertChain, and kSecImportItemIdentity Pkcs12ImportOptions::import will panic with a message such as:

"No entry found for key 0x1ab4c9fd0"

Because of this we can't open pkcs12 file created using openssl. (Other might work)

I believe not all of these keys are required.

PR #38, opened 2017-06-02
Handle missing 'items' from pkcs12 file more gracefully

azdlowry:
This will either use a default when possible, or will give a more useful error message should particular 'items' be missing from pkcs12 file.

See #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,4 @@ at your option.
 Unless you explicitly state otherwise, any contribution intentionally submitted
 for inclusion in the work by you shall be dual licensed as above, without any
 additional terms or conditions.
+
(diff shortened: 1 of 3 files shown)


sfackler:
I think it might make more sense to deprecate the existing method and struct and make a new one with optional fields. Does that seem reasonable?

azdlowry:
It does sound reasonable, but it might not be something I can do as I don't have access to a dev environment that can build the code. I'll see what I can do with my current pushing to travis.ci approach.

Also I'd probably still recommend keeping the error handling changes as the current error message just mentions a missing key, without mentioning which key is missing.

sfackler:
Looks good, thanks!

PR #41, opened 2017-06-28
support iOS

king6cong:


sfackler:
The portions of the library that are `#[cfg(target_os = "macos")]` are like that because those features do not exist on iOS, as the iOS Travis build indicates: https://travis-ci.org/sfackler/rust-security-framework/jobs/247854494

PR #42, opened 2017-08-04
upgrade core-foundation dep version

cjubb39:
I know you just cut a build yesterday.  Are you willing to cut another with just an upgrade in the dep version for `core-foundation`?  This will help me get rid of some duplicate deps in my project.

It's a clean update since core-foundation is very unlikely to have a breaking change :)

Thanks!

### DIFF
--- a/security-framework-sys/Cargo.toml
+++ b/security-framework-sys/Cargo.toml
@@ -12,7 +12,7 @@ build = "build.rs"
 
 [dependencies]
 libc = "0.2"
-core-foundation-sys = "0.2.1"
+core-foundation-sys = "0.4"
 
 [features]
 OSX_10_8 = []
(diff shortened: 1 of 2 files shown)


cjubb39:
Can also update the version in the `Cargo.toml` to "*" if you'd prefer

sfackler:
Wildcard dependencies can't be published.

This will be a breaking change, so I'm going to fold some other fixes into it.

cjubb39:
Why is this a breaking change?  There were no code changes to either of the crates in this repo.

sfackler:
core-foundation 0.2's types are not the same as core-foundation 0.4's types, so anyone that depends on core-foundation 0.2 and security-framework 0.1 is going to be broken if the dependency is bumped.

jrmuizel:
core-foundation has been updated so this can be closed.

PR #44, opened 2017-08-12
Fix incorrect type in certificate_at_index()

### LINKED ISSUE #43: i686 mismatched types
Getting an issue on travis, compiling for `i686-apple-darwin`

```
error[E0308]: mismatched types

   --> /Users/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/security-framework-0.1.15/src/trust.rs:124:69

    |

124 |             let certificate = SecTrustGetCertificateAtIndex(self.0, ix);

    |                                                                     ^^ expected i32, found i64
```

mattico:
Fixes #43
Also adds a single i686 travis build, so this type of thing can be caught in the future. (Also I don't have any way to test this besides travis)

### DIFF
--- a/security-framework/src/trust.rs
+++ b/security-framework/src/trust.rs
@@ -119,7 +119,7 @@ impl SecTrust {
     /// Returns a specific certificate from the certificate chain used to evaluate trust.
     ///
     /// Note: evaluate must first be called on the SecTrust.
-    pub fn certificate_at_index(&self, ix: i64) -> Option<SecCertificate> {
+    pub fn certificate_at_index(&self, ix: CFIndex) -> Option<SecCertificate> {
         unsafe {
             let certificate = SecTrustGetCertificateAtIndex(self.0, ix);
             if certificate.is_null() {
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,12 @@ matrix:
       - osx_image: xcode7.3
         env: TARGET=x86_64-apple-ios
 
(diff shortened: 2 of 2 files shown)


security-framework/src/trust.rs
```diff
@@ -119,9 +119,9 @@ impl SecTrust {
     /// Returns a specific certificate from the certificate chain used to evaluate trust.
     ///
     /// Note: evaluate must first be called on the SecTrust.
-    pub fn certificate_at_index(&self, ix: i64) -> Option<SecCertificate> {
+    pub fn certificate_at_index(&self, ix: isize) -> Option<SecCertificate> {
         unsafe {
-            let certificate = SecTrustGetCertificateAtIndex(self.0, ix);
+            let certificate = SecTrustGetCertificateAtIndex(self.0, ix as CFIndex);
```
mattico: This is a no-op since CFIndex = i32 on 32-bit and i64 on 64-bit.

mattico:
This is a breaking change, I suppose. May just want to publish 0.1.16 and yank 0.1.15.

security-framework/src/trust.rs
sfackler: Why not use CFIndex in the signature directly then?

mattico: No particular reason. Thought it was desirable to use 'rusty' types in the signature. I'll change it.

jchien14:
Ah good catch. Sorry about that :/

PR #46, opened 2017-09-03
Add option to allow connections with invalid ssl certificates

scottschroeder:
## Motivation

I know this is just... _wrong_, and everyone should be using letsencrypt, or adding their self-signed cert as a trusted root. Allowing invalid SSL connections is still a "feature" that rust doesn't have, and I'd like to fix that.

## Changes

This change adds an optional flag onto the `ClientBuilder` to accept invalid certificates.

#### Previously
The validation logic treated the `RecoverableTrustFailure` identically to the `FatalTrustFailure`. Upon a limited reading of the [documentation](https://developer.apple.com/documentation/security/sectrustresulttype/1397633-recoverabletrustfailure), it seems as though the intended use of the `RecoverableTrustFailure` was to allow the user to opt into the insecure connection.

#### Now
During validation, if we get a `RecoverableTrustFailure`, and the user has opted into allowing invalid certs, I treat it just like a `Proceed`. If the user has _not_ opted-in, the result is the same as it was before.

In keeping with rust's theme of not wanting you to shoot yourself in the foot, I've named the function `danger_accept_invalid_certs()`, and put a big warning in the doc comments. If you want this to be more obnoxious name, I'm open to suggestions.

## Testing

I added a test-case from [badssl.com](https://badssl.com). I also tested this by making connections to a variety of bad servers from my osx workstation.

### DIFF
--- a/security-framework/src/secure_transport.rs
+++ b/security-framework/src/secure_transport.rs
@@ -222,6 +222,7 @@ pub struct MidHandshakeClientBuilder<S> {
     domain: Option<String>,
     certs: Vec<SecCertificate>,
     trust_certs_only: bool,
+    danger_accept_invalid_certs: bool,
 }
 
 impl<S> MidHandshakeClientBuilder<S> {
@@ -242,7 +243,14 @@ impl<S> MidHandshakeClientBuilder<S> {
 
     /// Restarts the handshake process.
     pub fn handshake(self) -> result::Result<SslStream<S>, ClientHandshakeError<S>> {
-        let MidHandshakeClientBuilder { stream, domain, certs, trust_certs_only } = self;
+        let MidHandshakeClientBuilder {
+            stream,
+            domain,
+            certs,
+            trust_certs_only,
+            danger_accept_invalid_certs
+        } = self;
+
         let mut result = stream.handshake();
         loop {
             let stream = match result {
@@ -257,6 +265,7 @@ impl<S> MidHandshakeClientBuilder<S> {
                     domain: domain,
                     certs: certs,
                     trust_certs_only: trust_certs_only,
+                    danger_accept_invalid_certs: danger_accept_invalid_certs,
                 };
                 return Err(ClientHandshakeError::Interrupted(ret));
             }
@@ -281,7 +290,15 @@ impl<S> MidHandshakeClientBuilder<S> {
                         let err = Error::from_code(errSecTrustSettingDeny);
                         return Err(ClientHandshakeError::Failure(err));
                     }
-                    TrustResult::RecoverableTrustFailure |
+                    TrustResult::RecoverableTrustFailure => {
+                        if danger_accept_invalid_certs {
+                            result = stream.handshake();
+                            continue;
+                        } else {
+                            let err = Error::from_code(errSecNotTrusted);
+                            return Err(ClientHandshakeError::Failure(err));
+                        }
+                    }
                     TrustResult::FatalTrustFailure => {
                         let err = Error::from_code(errSecNotTrusted);
                         return Err(ClientHandshakeError::Failure(err));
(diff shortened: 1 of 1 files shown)


security-framework/src/secure_transport.rs
```diff
@@ -281,7 +290,15 @@ impl<S> MidHandshakeClientBuilder<S> {
                         let err = Error::from_code(errSecTrustSettingDeny);
                         return Err(ClientHandshakeError::Failure(err));
                     }
-                    TrustResult::RecoverableTrustFailure |
+                    TrustResult::RecoverableTrustFailure => {
```
sfackler: You can simplify this to `TrustResult::RecoverableTrustFailure if danger_accept_invalid_certs => { ... }` and then keep the arm below unmodified to avoid having to duplicate the not trusted stuff.

security-framework/src/secure_transport.rs
```diff
@@ -1386,6 +1419,15 @@ mod test {
     }
 
     #[test]
+    fn connect_no_verify_ssl() {
+        let stream = p!(TcpStream::connect("expired.badssl.com:443"));
+        let mut builder = ClientBuilder::new();
+        builder.danger_accept_invalid_certs(true);
+        builder.danger_handshake_without_providing_domain_for_certificate_validation_and_server_name_indication(stream)
```
sfackler: It should be fine to use the normal handshake method here, right?

sfackler:
Sorry for missing this for a while!

security-framework/src/secure_transport.rs
scottschroeder: Yes, that's much nicer looking, thanks!

scottschroeder:
I updated this PR based on your feedback. Let me know if there's anything else I can do.

Once you decide to push up a new crate with this change, I'll update my PR for https://github.com/sfackler/rust-native-tls/pull/56, and we can go from there.

security-framework/src/secure_transport.rs
```diff
@@ -281,6 +291,10 @@ impl<S> MidHandshakeClientBuilder<S> {
                         let err = Error::from_code(errSecTrustSettingDeny);
                         return Err(ClientHandshakeError::Failure(err));
                     }
+                    TrustResult::RecoverableTrustFailure if danger_accept_invalid_certs => {
```
sfackler: Actually, it seems like we arguably would want to skip the trust check entirely if we have danger_accept_invalid_certs set, right?

scottschroeder: Honestly, I'm not really sure. According to the [apple docs](https://developer.apple.com/documentation/security/sectrustresulttype) the `RecoverableTrustFailure` is the only one that we (as a piece of software) should be able to make a judgement call on. But that might be more directed at a high level app, and less at a low-level lib. If I explicitly decided to [deny](https://developer.apple.com/documentation/security/sectrustresulttype/1400766-deny) a cert chain, should this lib respect that?

Given that this will probably be used from `rust-native-tls`, I doubt we can expose these nuances meaningfully. Furthermore, I think the linux/windows implementations from my PR there will blanket accept whatever.

I think I agree with you then, that we should skip the check entirely.

scottschroeder:
Sorry, I pushed a little early, I don't trust what I just wrote, seems like we could either get stuck in that loop, or error out with the wrong message.

scottschroeder:
Ok, I think its about as sane as we're going to get. I think worse case scenario would be some kind of corrupted cert which is both invalid, and unable to even support SSL. We would "trust" it, but I think it would end up causing some kind of handshake or other stream error.

Let me know if you can think of anything else.

sfackler:
Yeah - my intuition is that if you're turning off certificate verification, you're doing it because you trust whatever garbage the server's giving to you (!!!), so we'd want to turn everything off.

PR #48, opened 2018-01-27
Update to core-foundation 0.5

jrmuizel:
It sounded like a version bump was coming, so I didn't include that in this PR


sfackler:
It looks like the Rust version used on Travis needs to be bumped.

jrmuizel:
This passes CI now

PR #49, opened 2018-03-19
Ensure correct FFI bindings

sfackler:


sfackler:
Some complications here: `Uint32` changed from `unsigned long` to `unsigned int` a couple of SDKs ago, and `SecCopyRandomByte`'s last argument changed from `unsigned char *` to `void *`.

PR #56, opened 2018-05-01
Add ALPN bindings to security-framework-sys

lqf96:
This pull request adds ALPN function bindings to the crate.

The bindings are currently in module "secure_transport_priv" because these functions are defined in [SecureTransportPriv.h](https://opensource.apple.com/source/Security/Security-58286.41.2/OSX/libsecurity_ssl/lib/SecureTransportPriv.h.auto.html). Because these functions are currently private APIs, I extracted the API definitions and put them in a header file with the same name under "priv-headers" so that the binding tests can continue to work.

PR #57, opened 2018-05-02
Add ALPN bindings to security-framework-sys, take 2

lqf96:
Follow up of #56.

This pull request adds ALPN bindings for function `SSLSetALPNProtocols` and `SSLCopyALPNProtocols`, both of which are publicly documented and only available for macOS 10.13+. Binding tests for macOS 10.12/10.13 are also added.


sfackler:
Can you forward this through the SslContext? This may have been fixed, but one or both of those functions wasn't actually present in Security.framework.

lqf96:
As of macOS 10.13.4 both of the functions are indeed in `Security.framework`, but I do hear that `
SSLSetALPNProtocols` was not present in some initial versions of macOS 10.13.



As for the high-level ALPN APIs in `SslContext`, I plan to do it in another pull request.

lqf96:
I added `alpn_protocols` and `set_alpn_protocols` to `SslContext` since it seems more appropriate to make them available in this pull request. Do you have any problems with the function signatures? Also, what should we do about the missing symbol bug in initial macOS 10.13 versions?

sfackler:
Sorry for the delay! I updated the branch to fix the merge conflicts, and it looks good to me!

PR #62, opened 2018-06-06
Some ALPN tweaks

sfackler:


security-framework/src/secure_transport.rs
```diff
@@ -699,31 +700,35 @@ impl SslContext {
...
+            Ok(array.into_iter().map(|p| p.to_string()).collect())
+        }
     }
 
-    /// Sets the Application Layer Protocol Negotiation (ALPN) protocols list.
+    /// Configures the set of protocols use for ALPN.
+    ///
+    /// This is only used for client-side connections.
     #[cfg(feature = "OSX_10_13")]
     pub fn set_alpn_protocols(&mut self, protocols: &[&str]) -> Result<()> {
```
lqf96: I'm wondering if it's better to change the signature of `protocols` to `&[String]`? This way we can avoid creating a temporary vector in `ClientBuilder`

sfackler: Yeah I was going back and forth with that. The nice part about `&[&str]` is that it's almost always going to be called with static string literals like `ctx.set_alpn_protocols(&["h2", "http/1.1"])` and it'd be nice to not have to do `ctx.set_alpn_protocols(&["h2".to_string(), "http/1.1".to_string()])`. We could just make it generic over `T: AsRef<str>`, but that's not really a pattern we've used elsewhere in the library.

PR #66, opened 2018-07-15
Add bindings for password functions

kornelski:
The higher-level wrapper for passwords could be more ergonomic, so I've removed it from the PR. This is just the uncontroversial `-sys` bindings part.


kornelski:
Ping @sfackler

kornelski:
Ping. This is a simple addition.

PR #67, opened 2018-07-18
Higher-level wrapper for passwords

kornelski:
Based on #17, with additions:

* Supports "Internet" passwords in addition to generic app passwords.
* Adds methods to `SecKeychainItem` instead of using standalone functions, which removes some repetition from the code.
* Avoids creating temporary `CString`s, since Keychain supports explicit string lengths.
* Includes examples.


kornelski:
I'd love more feedback about these APIs. Let me know if you use them.

PR #69, opened 2018-08-09
Improve Keychain Item SearchResult(s).

archaelus:
This PR is another take on generic password support for security-framework:
* ItemSearchOptions can now return item attributes and data (passwords for
internet and generic password item classes).
* SearchResult(s) are now an enum depending on the return kind of the
search. CFDict results can be simplified to a HashMap<String, String>.

This work in turn allows callers to retrieve passwords and associated data (like an account url) from the keychain:
```
    let results = ItemSearchOptions::new()
        .class(ItemClass::generic_password())
        .label(&args.label)
        .load_refs(true)
        .load_attributes(true)
        .load_data(true)
        .search()?;

    let result = results
        .first()
        .ok_or(format_err!("No results for that keychain label."))?;
    let mut dict = result
        .simplify_dict() // I added this because the raw keychain result is pretty hard to work with.
        .ok_or(format_err!("Couldn't interpret keychain item."))?;
    let acct: String = dict
        .remove("acct")
        .ok_or(format_err!("No account info in the keychain item."))?;
    let pass: String = dict
        .remove("v_Data")
        .ok_or(format_err!("No password info in the keychain item."))?;

    println!("Got account: {}, pass: {}", acct, pass);
```

Being able to return both account URL and password by searching for a label makes writing CLI utilities much more convenient. A user can store their api key for some service in the keychain by creating a generic password or internet password, and the CLI utility can look up a supplied label, and get back all the data (service URL) and credentials it needs to access the service.

### DIFF
--- a/security-framework/src/item.rs
+++ b/security-framework/src/item.rs
@@ -3,12 +3,15 @@
 use core_foundation::array::CFArray;
 use core_foundation::base::{CFType, TCFType};
 use core_foundation::boolean::CFBoolean;
+use core_foundation::data::CFData;
+use core_foundation::date::CFDate;
 use core_foundation::dictionary::CFDictionary;
 use core_foundation::number::CFNumber;
 use core_foundation::string::CFString;
-use core_foundation_sys::base::{CFGetTypeID, CFRelease, CFTypeRef};
+use core_foundation_sys::base::{CFCopyDescription, CFGetTypeID, CFRelease, CFTypeRef};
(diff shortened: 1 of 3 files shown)


kornelski:
Hi! Sorry for the delay.

Is this PR still good to merge?

archaelus:
I think so, yes - this patch could use docs and testing, but the functionality still builds and works afaict. It was really useful when building CLI utilities that stored creds in the keychain, so I think it's still a good idea to merge it.

PR #72, opened 2019-01-23
Remove dependency on MacTypes-sys

luben:
MacTypes-sys APSL-2.0 license is preventing this library use
in a large number of contexts.


kornelski:
Thanks

luben:
Thanks!

PR #73, opened 2019-02-06
additional exports and wrappers

SergejJurecko:


security-framework/src/trust.rs
```diff
@@ -103,6 +104,11 @@ impl SecTrust {
         unsafe { cvt(SecTrustSetPolicies(self.0, policy.as_CFTypeRef())) }
     }
 
+    /// Returns the public key for a leaf certificate after it has been evaluated.
+    pub fn copy_public_key(&mut self) -> Result<SecKey> {
+        unsafe { Ok(SecKey::wrap_under_get_rule(SecTrustCopyPublicKey(self.0))) }
```
kornelski: Are you sure about the get rule? I think `Copy` functions behave like `Create` ones.

SergejJurecko: Honestly I do not know. We can remove this for now perhaps and leave the low level APIs?

kornelski: It's fine. Just change it to create rule.

SergejJurecko: Done

kornelski:
Thank you

PR #75, opened 2019-02-13
SecCertificate::public_key_info_der 

SergejJurecko:


security-framework-sys/src/key.rs
```diff
@@ -1,8 +1,8 @@
 use core_foundation_sys::base::CFTypeID;
-#[cfg(target_os = "macos")]
+#[cfg(any(target_os = "macos", target_os = "ios"))]
 use core_foundation_sys::data::CFDataRef;
 use core_foundation_sys::dictionary::CFDictionaryRef;
-#[cfg(target_os = "macos")]
```
kornelski: if it's on both targets, then the cfg can be removed.

security-framework/src/certificate.rs
```diff
@@ -57,8 +86,152 @@ impl SecCertificate {
             CFString::wrap_under_create_rule(summary).to_string()
         }
     }
+
+    #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
+    /// Returns DER encoded subjectPublicKeyInfo of certificate if available. This can be used
+    /// for certificate pinning.
+    pub fn public_key_info_der(&self) -> Result<Option<Vec<u8>>> {
```
kornelski: My concern is that this is a very specific single-purpose high-level function. Could it be broken down into components that help others write their `public_key_info_der` and alike themselves?

security-framework/src/certificate.rs
```diff
@@ -57,8 +86,152 @@ impl SecCertificate {
...
+        // Imported from TrustKit
+        // https://github.com/datatheorem/TrustKit/blob/master/TrustKit/Pinning/TSKSPKIHashCache.m
+        unsafe {
+            let public_key = self.copy_public_key_from_certificate()?;
```
kornelski: Is there a reason why this function is not public?

security-framework/src/certificate.rs
```diff
@@ -57,8 +86,152 @@ impl SecCertificate {
...
+            let mut error: CFErrorRef = ptr::null_mut();
+            let public_key_attributes = SecKeyCopyAttributes(public_key);
```
kornelski: Maybe there could be a high-level wrapper for `SecKeyCopyAttributes`, so that it doesn't require so much unsafe boilerplate for each property?

security-framework/src/certificate.rs
```diff
@@ -57,8 +86,152 @@ impl SecCertificate {
...
+            return &RSA_4096_ASN1_HEADER;
+        }
+        if CFStringCompare(pkt, kSecAttrKeyTypeECSECPrimeRandom, 0) as i64 == 0 && ksz == 256 {
+            return &EC_DSA_SECP_256_R1_ASN1_HEADER;
+        }
+        if CFStringCompare(pkt, kSecAttrKeyTypeECSECPrimeRandom, 0) as i64 == 0 && ksz == 384 {
+            return &EC_DSA_SECP_384_R1_ASN1_HEADER;
+        }
+    }
+    &[]
```
kornelski: Better error handling would be to return `None`

security-framework/src/certificate.rs
```diff
@@ -57,8 +86,152 @@ impl SecCertificate {
...
+}
+
+extern "C" {
+    #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
+    pub fn CFStringCompare(
```
kornelski: You don't need this. `CFString` implements `Eq`, so you can compare them with just `==`

SergejJurecko: I'm using CFStringRef's due to comparing with constants kSecAttrKeyTypeRSA and kSecAttrKeyTypeECSECPrimeRandom. Can I turn those into CFString without allocating?

SergejJurecko:
I've expanded SecKey and cut down on the big unsafe block into mostly safe code.

security-framework/src/certificate.rs
kornelski: I think you can. The `CFString` is a refcounted type, so a conversion from `Ref` to non-Ref should only need to change reference count:

https://docs.rs/core-foundation/0.6.3/core_foundation/base/trait.TCFType.html#tymethod.wrap_under_get_rule

kornelski: Also the eq is implemented as `self.as_CFType().eq(&other.as_CFType())`

security-framework/src/certificate.rs
```diff
@@ -57,8 +83,124 @@ impl SecCertificate {
...
+    #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
+    /// Returns DER encoded subjectPublicKeyInfo of certificate if available. This can be used
+    /// for certificate pinning.
+    pub fn public_key_info_der(&self) -> Result<Option<Vec<u8>>> {
+        // Imported from TrustKit
+        // https://github.com/datatheorem/TrustKit/blob/master/TrustKit/Pinning/TSKSPKIHashCache.m
+        let public_key = self.public_key()?;
+        let public_key_attributes = public_key.attributes();
+        let public_key_type =
+            public_key_attributes.find(unsafe { kSecAttrKeyType } as *const std::os::raw::c_void);
```
kornelski: `kSecAttrKeyType` is a constant right? So `unsafe { kSecAttrKeyType }` shouldn't need unsafe.

SergejJurecko: It's apparently a new thing by the compiler:

error: use of extern static is unsafe and requires unsafe function or block (error E0133)
 note: #[deny(safe_extern_statics)] on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!

security-framework/src/certificate.rs
SergejJurecko: Ok switched to CFString now.

security-framework/src/certificate.rs
kornelski: Ah, right. It's not a (Rust) constant, but technically a global mutable variable. So unsafe makes sense!

SergejJurecko:
Another small refactor to get rid of verbose error handling. Do you have any other suggestions for improving it?

kornelski:
This is much nicer now. Thank you!

SergejJurecko:
Can this be released?

PR #77, opened 2019-02-28
Fix for older rustc

SergejJurecko:


SergejJurecko:
Sorry to open this again. A slight fix for older rustc versions is required so travis stops complaining.

kornelski:
Thanks

PR #82, opened 2019-09-13
Add SecTrustSettings bindings

ctz:
This allows a consuming crate to enumerate macos trusted certificates via `SecTrustSettingsCopyCertificates`, and then learn whether they're suitable for use in TLS via `SecTrustSettingsCopyTrustSettings`.

I've done some manual testing of this by twiddling keychain trust settings and observing the results.

Sample calling code is here if you're interested: https://github.com/ctz/rustls-native-certs/blob/master/src/macos.rs

### DIFF
--- a/security-framework/src/trust_settings.rs
+++ b/security-framework/src/trust_settings.rs
@@ -117,8 +117,16 @@ impl TrustSettings {
     ///
     /// This tells you whether the certificate should be trusted as a TLS
     /// root certificate.
+    ///
+    /// If the certificate has no trust settings in the given domain, the
+    /// `errSecItemNotFound` error is returned.
+    ///
+    /// If the certificate has no specific trust settings for TLS in the
+    /// given domain `None` is returned.
+    ///
+    /// Otherwise, the specific trust settings are aggregated and returned.
     pub fn tls_trust_settings_for_certificate(&self, cert: &SecCertificate)
-        -> Result<TrustSettingsForCertificate> {
+        -> Result<Option<TrustSettingsForCertificate>> {
         let trust_settings = unsafe {
             let mut array_ptr: CFArrayRef = ptr::null_mut();
             let cert_ptr = cert.as_CFTypeRef() as *mut _;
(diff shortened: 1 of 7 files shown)


security-framework/src/trust_settings.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    #[test]
+    fn test_list_for_user() {
+        list_for_domain(Domain::User);
+    }
+
+    #[test]
+    fn test_list_for_system() {
+        list_for_domain(Domain::System);
```
kornelski: These aren't real tests. Could you check for some known cert? (some longer-lived one that macOS ships with. It's ok if the test breaks in a few years).

security-framework/src/trust_settings.rs
```diff
@@ -0,0 +1,253 @@
...
+            // that to an empty TrustSettings iterator.
+            match SecTrustSettingsCopyCertificates(self.domain.into(), &mut array_ptr) {
+                errSecNoTrustSettings => {
+                    Ok(CFArray::from_CFTypes(&[]))
+                },
+                errSecSuccess => {
+                    Ok(CFArray::<SecCertificate>::wrap_under_create_rule(array_ptr))
+                },
+                err => Err(Error::from_code(err)),
+            }?
```
kornelski: nit: you can add `?` to `Err` arm directly, so the others won't need to be Ok-wrapped.

security-framework/src/trust_settings.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// This tells you whether the certificate should be trusted as a TLS
+    /// root certificate.
+    pub fn tls_trust_settings_for_certificate(&self, cert: &SecCertificate)
+        -> Result<TrustSettingsForCertificate> {
+        let trust_settings = unsafe {
+            let mut array_ptr: CFArrayRef = ptr::null_mut();
+            let cert_ptr = cert.as_CFTypeRef() as *mut _;
+            match SecTrustSettingsCopyTrustSettings(cert_ptr, self.domain.into(), &mut array_ptr) {
+                errSecNoTrustSettings => {
+                    Ok(CFArray::from_CFTypes(&[]))
```
kornelski: The docs say " An empty Trust Settings array is definitely not the same as *no* Trust Settings, which means "this cert must be verified to a known trusted cert". , but an empty array here will fall back to trust!

security-framework/src/trust_settings.rs
ctz: Good idea. I've added one of these for the Let's Encrypt/ISRG root.

security-framework/src/trust_settings.rs
ctz: Yeah, it seems reasonable to return None here if there were no errors but, equally, no specific settings. I'll do that.

ctz:
Thanks for the time you spent reviewing this. I think I've addressed all your comments in some remedial commits; let me know if you prefer these to be squashed.

kornelski:
Great. Thank you!

PR #83, opened 2019-10-15
Use a dictionary with string keys for the result of SecPKCS12Import.

jrmuizel:
This adds some extra typesafety.


kornelski:
Thanks

PR #91, opened 2019-11-28
Add Authorization Services support

steven-joruk:
This is my first time writing Rust FFI bindings, I'm happy to make as many changes as are needed.

~I've kept in some documented potential panics, but they can only be reached due to user error in forming strings which can't be converted to CString. I could instead map the CString::new error to a suitable OSStatus, like errSecInvalidData or errSecInvalidEncoding.~ CString::new failures map to errSecConversionError.

AuthorizationExecuteWithPrivileges has been deprecated for years now, but is still usable in Catalina. Adding the deprecated attribute causes extra warning noise in systest.

I went through a few iterations of how to implement the AuthorizationItemSetBuilder and settled on this. I expect it would be a bit simpler using alloc/dealloc, but wanted to avoid manual memory management. It's important that AuthorizationItemSetStorage's names and values fields aren't modified after the builder is consumed, and it's safe from that from a user's perspective.

Some test are ignored, because they require an admin user's name and password. I can look in to updating the CI and enabling these tests if you like.

### DIFF
--- a/security-framework-sys/src/authorization.rs
+++ b/security-framework-sys/src/authorization.rs
@@ -1,0 +1,145 @@
+use core_foundation_sys::base::CFTypeRef;
+use core_foundation_sys::base::OSStatus;
+use core_foundation_sys::bundle::CFBundleRef;
+use core_foundation_sys::dictionary::CFDictionaryRef;
+use core_foundation_sys::string::CFStringRef;
+use std::os::raw::{c_char, c_void};
+
+pub const errAuthorizationSuccess: OSStatus = 0;
+pub const errAuthorizationInvalidSet: OSStatus = -60001;
+pub const errAuthorizationInvalidRef: OSStatus = -60002;
+pub const errAuthorizationInvalidTag: OSStatus = -60003;
+pub const errAuthorizationInvalidPointer: OSStatus = -60004;
+pub const errAuthorizationDenied: OSStatus = -60005;
+pub const errAuthorizationCanceled: OSStatus = -60006;
+pub const errAuthorizationInteractionNotAllowed: OSStatus = -60007;
+pub const errAuthorizationInternal: OSStatus = -60008;
+pub const errAuthorizationExternalizeNotAllowed: OSStatus = -60009;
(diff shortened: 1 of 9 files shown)


kornelski:
Thanks for the PR. Sorry I haven't had time to review it yet.

steven-joruk:
There’s no rush. I would like to add another couple of tests and try to run it through valgrind anyway.

security-framework-sys/src/authorization.rs
```diff
@@ -0,0 +1,150 @@
...
+
+pub type AuthorizationFlags = u32;
+pub const kAuthorizationFlagDefaults: AuthorizationFlags = 0;
+pub const kAuthorizationFlagInteractionAllowed: AuthorizationFlags = 1;
+pub const kAuthorizationFlagExtendRights: AuthorizationFlags = 2;
+pub const kAuthorizationFlagPartialRights: AuthorizationFlags = 4;
+pub const kAuthorizationFlagDestroyRights: AuthorizationFlags = 8;
+pub const kAuthorizationFlagPreAuthorize: AuthorizationFlags = 16;
+
+pub type AuthorizationRef = *mut c_void;
```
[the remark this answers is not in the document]
steven-joruk: I see a crash in the tests when I introduce this change, the other two are in though.

security-framework/src/authorization.rs
```diff
@@ -0,0 +1,641 @@
...
+        let status =
+            unsafe { sys::AuthorizationCopyInfo(self.handle, tag_ptr, inner.as_mut_ptr()) };
+
+        if status != sys::errAuthorizationSuccess {
+            return Err(Error::from(status));
+        }
+
+        let set = AuthorizationItemSet {
+            inner: unsafe { inner.assume_init() },
+            is_owned: false,
```
kornelski: Why isn't it owned? `Copy` functions usually return a new object (or recount++ object).

steven-joruk: Thanks for spotting that, I missed this during refactoring - is_owned can be removed and Drop for it should always call AuthorizationFreeItemSet.

PR #93, opened 2020-01-12
Expose kTLSProtocol13 for TLS 1.3

mehcode:
https://developer.apple.com/documentation/security/sslprotocol/ktlsprotocol13?language=objc

macOS 10.14+
iOS 12.2+


kornelski:
Thanks

kornelski:
This PR just added a constant. No guarantees about Apple's implementation :)

PR #95, opened 2020-03-11
Support TLS session ticket resumption

### LINKED ISSUE #94: Add an option to support TLS session ticket resumption
From some manual testing (by connecting to a local server run with `openssl s_server -key key.pem -cert cert.pem -accept 44330 -tlsextdebug -www -state`), it looks like there are two requirements here:
  - Call either `SSLSetSessionTicketsEnabled` or `SSLSetSessionOption` with `kSSLSessionOptionEnableSessionTickets`, which will cause the client to send the `SessionTicket` extension in its initial `ClientHello`.
  - Call `SSLSetPeerID` with something that uniquely identifies the remote (maybe domain + port?).
Both of these are required, as with just the first, the client will just send an empty `SessionTicket` extension in its `ClientHello` every time. It seems like setting a peer ID is required so that Security Framework knows what to key the session ticket against.

In the `openssl s_server` output, a normal TLS handshake looks like:
```
SSL_accept:before/accept initialization
TLS client extension "session ticket" (id=35), len=0
...
SSL_accept:SSLv3 read client hello A
SSL_accept:SSLv3 write server hello A
SSL_accept:SSLv3 write certificate A
SSL_accept:SSLv3 write key exchange A
SSL_accept:SSLv3 write server done A
SSL_accept:SSLv3 flush data
SSL_accept:SSLv3 read client certificate A
SSL_accept:SSLv3 read client key exchange A
SSL_accept:SSLv3 read certificate verify A
SSL_accept:SSLv3 read finished A
SSL_accept:SSLv3 write session ticket A
SSL_accept:SSLv3 write change cipher spec A
SSL_accept:SSLv3 write finished A
SSL_accept:SSLv3 flush data
```
and a handshake with TLS session ticket resumption looks like:
```
SSL_accept:before/accept initialization
TLS client extension "session ticket" (id=35), len=176
...
SSL_accept:SSLv3 read client hello A
SSL_accept:SSLv3 write server hello A
SSL_accept:SSLv3 write change cipher spec A
SSL_accept:SSLv3 write finished A
SSL_accept:SSLv3 flush data
SSL_accept:SSLv3 read finished A
```

cyang1:
Initial draft of session ticket resumption support.

There's a bit that probably needs to be fleshed out here:
  - Like ALPN, the relevant APIs for session ticket resumption are also only available on 10.13 and above. I'm currently handling this in a similar way to ALPN by introducing a new feature, but I'm not sure this is a good idea/scalable way to expose features only available on newer versions.
  - The public API for this is a bit convoluted, `pub fn enable_session_tickets(&mut self, address: &[u8]) -> &mut Self`. This is because we need a parameter to pass to `set_peer_id`, and AFAICT, a valid parameter isn't available to us at this abstraction level, without the user passing it in manually.

Fixes #94.

### DIFF
--- a/security-framework-sys/src/secure_transport.rs
+++ b/security-framework-sys/src/secure_transport.rs
@@ -266,4 +266,6 @@ extern "C" {
     pub fn SSLSetALPNProtocols(context: SSLContextRef, protocols: CFArrayRef) -> OSStatus;
     #[cfg(feature = "OSX_10_13")]
     pub fn SSLCopyALPNProtocols(context: SSLContextRef, protocols: *mut CFArrayRef) -> OSStatus;
+    #[cfg(feature = "OSX_10_13")]
+    pub fn SSLSetSessionTicketsEnabled(context: SSLContextRef, enabled: Boolean) -> OSStatus;
 }
--- a/security-framework/Cargo.toml
+++ b/security-framework/Cargo.toml
@@ -24,12 +24,13 @@ hex = "0.4"
 
 [features]
 alpn = []
+session-tickets = []
 
 OSX_10_9 = ["security-framework-sys/OSX_10_9"]
 OSX_10_10 = ["OSX_10_9", "security-framework-sys/OSX_10_10"]
 OSX_10_11 = ["OSX_10_10", "security-framework-sys/OSX_10_11"]
 OSX_10_12 = ["OSX_10_11", "security-framework-sys/OSX_10_12"]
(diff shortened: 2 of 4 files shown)


cyang1:
On second thought, I suspect we don't need the peer address throughout, as TLS session ticket resumption is meant to be stateless and work across different server as long as they share the same key.

kornelski:
That looks good. Thank you.

PR #97, opened 2020-03-30
Add a conditional `extern crate libc`

### LINKED ISSUE #96: dlsym.rs is missing a libc dependency
Trying to use this crate with either `alpn` or `session-tickets` and _not_ specifying `OSX_10_13` results in:
```
error[E0432]: unresolved import `libc`
 --> security-framework-0.4.2-alpha.1/src/dlsym.rs:8:5
  |
8 | use libc;
  |     ^^^^ no `libc` in the root
```

cyang1:
Fixes #96


kornelski:
Thanks.

PR #99, opened 2020-05-05
Rust 2018 fixes for < OSX_10_13 and ALPN

cyang1:
Referring to crate root is slightly different in Rust 2018, and it looks like raw pointers need a type specified.


kornelski:
Thank you

PR #102, opened 2020-05-12
Fix examples to work on macOS.

ehuss:
It seems like the `target_os` was set to an invalid value for the examples. Updating this makes them work for my mac system.


kornelski:
Thanks

PR #110, opened 2021-01-22
Implement TryFrom<AuthorizationExternalForm>

steven-joruk:
This fixes a TODO I left in from when it was originally accepted. I've deprecated `from_external_form` "from 2.0.1", but that's a bit speculative.


kornelski:
Thank you

PR #111, opened 2021-01-22
Set up GitHub Actions

steven-joruk:
Closes #5 

I've just hardcoded 1.34.0, but if the minimum supported version is formalised it would be good to make use of whatever it is automatically.

This fails at the moment, but I think the problems would be better fixed in follow up PRs:
* The formatting for 1.34.0 needs fixing
* The clippy warnings for 1.34.0 need fixing or ignoring for the bindings
* The 4 unit tests that fail need fixing or maybe there's some environment set up that I haven't seen the documentation for?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,58 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: master
+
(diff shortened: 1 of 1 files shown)


kornelski:
Thanks

PR #117, opened 2021-02-19
Add some more SecKey functionality

steven-joruk:
There's plenty more to add relating to SecKey. This at least enables me to verify that an item in the keychain is accessible by attempting to use it to create a signature while having user interaction disabled.

I tried to make the Algorithm enum a bit less horrific by using a macro, but the approach I attempted involved concat_ident which is a nightly only feature. I'll see if there's a nicer way of doing this if you don't know of one yourself.

### DIFF
--- a/security-framework/src/key.rs
+++ b/security-framework/src/key.rs
@@ -8,11 +8,15 @@ use core_foundation::data::CFData;
 #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
 use core_foundation::dictionary::CFDictionary;
 #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
-use core_foundation_sys::error::CFErrorRef;
+use core_foundation::error::{CFError, CFErrorRef};
 use security_framework_sys::base::SecKeyRef;
+#[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
+pub use security_framework_sys::key::Algorithm;
 use security_framework_sys::key::SecKeyGetTypeID;
 #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
-use security_framework_sys::key::{SecKeyCopyAttributes, SecKeyCopyExternalRepresentation};
+use security_framework_sys::key::{
+    SecKeyCopyAttributes, SecKeyCopyExternalRepresentation, SecKeyCreateSignature,
+};
 use std::fmt;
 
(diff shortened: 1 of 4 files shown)


security-framework/src/key.rs
```diff
@@ -24,6 +29,308 @@ impl_TCFType!(SecKey, SecKeyRef, SecKeyGetTypeID);
 unsafe impl Sync for SecKey {}
 unsafe impl Send for SecKey {}
 
+#[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
+#[allow(missing_docs)]
+/// The algorithms that cryptographic keys enable.
+pub enum Algorithm {
```
kornelski, line 35: I imagine Apple will add more algorithms in the future. Can you make this enum non_exhaustive?

kornelski: Also there names are repeated a few times. Maybe you could use a macro to reduce the boilerplate?

```rust
macro_rules! names {
    ($($i:ident => $x:ident),*) => {
    	extern "C" {
       		$(static $x: u32;)*
    	}
        enum Enum {
            $( $i ),*
        }
        fn from(m: Enum) -> u32 {
            unsafe { match m {
                $( $i => $x, )*
            } }
        }
    }
}

names! {
    foo => macFoo, bar => macBar, baz => macBaz
}
```

security-framework/src/os/macos/keychain.rs
```diff
@@ -74,6 +74,33 @@ impl SecKeychain {
             ))
         }
     }
+
+    #[cfg(target_os = "macos")]
+    /// Disables the user interface for keychain services functions that
+    /// automatically display a user interface.
+    pub fn disable_user_interaction() -> Result<KeychainUserInteractionLock> {
+        let code = unsafe { SecKeychainSetUserInteractionAllowed(0u8) };
```
kornelski, line 82: why not `set_user_interaction_allowed`?

steven-joruk:
@kornelski I'll update with your requests, cheers.

One thing I don't like is that this exposes the CFError, requiring users to add a direct dependency on core foundation. I looked at changing Error to an enum that contains either an OSStatus or CFError, but CFError is not Copy. Do you think it's a problem, and how could it be better?

kornelski:
I don't see a problem with exposing `CFError`. There already are other methods that expose it.

PR #125, opened 2021-03-08
Add Authorization::execute_with_privileges

ancwrd1:
This PR adds Authorization::execute_with_privileges function.

I know that it's considered deprecated but it is required for one of our projects and is actually quite useful so I thought to make a PR into the upstream. Otherwise it requires to fiddle around directly with security-framework-sys and low level API to call the `AuthorizationExecuteWithPrivileges`.

### DIFF
--- a/security-framework/src/authorization.rs
+++ b/security-framework/src/authorization.rs
@@ -536,11 +536,7 @@ impl<'a> Authorization {
             )
         };
 
-        if status != sys::errAuthorizationSuccess {
-            Err(Error::from(status))
-        } else {
-            Ok(())
-        }
+        crate::cvt(status)
     }
 }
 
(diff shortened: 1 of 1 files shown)


kornelski:
Thanks for the PR.

There's `cvt` function for making result from status that you could reuse.

How about using Rust's deprecated attribute on it to copy Apple's deprecation state?

ancwrd1:
> There's cvt function for making result from status that you could reuse.

Thanks, added it.

> How about using Rust's deprecated attribute on it to copy Apple's deprecation state?

I'd argue against it. Apple definition of "deprecated" is somewhat different in a sense that you are supposed to completely rewrite or redesign your code :)
In this particular case there is no easy replacement for this function and several complicated mechanisms are recommended:
https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/AccessControl.html

kornelski:
Thanks

PR #126, opened 2021-03-12
Add bindings for SecKeychainCopyDomainDefault

ddeville:
See https://developer.apple.com/documentation/security/1401993-seckeychaincopydomaindefault?language=objc.

This is super similar to `SecKeychainCopyDefault` that is used in `SecKeychain::default` but it takes an additional domain.


kornelski:
Thanks

PR #127, opened 2021-03-12
Fix missing re-export for `SecPreferencesDomain`

ddeville:
Forgot to re-export the symbol in #126, do that here so that users of `security-framework` don’t have to depend on `security-framework-sys` to use this function. Sorry for the noise.


kornelski:
Thanks

PR #129, opened 2021-04-19
New Item search features

kaisq:
This commit adds two new features to the Item search:

* Support for kSecMatchLimitAll. Previously, this library only allowed
searching for a single item, or for a specific number of items. This
feature allows all items available to be returned.

* Support for kSecAttrAccessGroup, specifically set to
kSecAttrAccessGroupToken. This allows retrieving items that have been
retrieved from a token, such as a smart card, via a CryptoTokenKit
provider.

Note that the current implementation for `kSecMatchLimitAll` is a breaking change. I think the new implementation is fairly clean and matches the feature for specifying the item class, but if you don't want a breaking change I'm happy to discuss alternatives.

### DIFF
--- a/security-framework/src/item.rs
+++ b/security-framework/src/item.rs
@@ -65,6 +65,32 @@ impl ItemClass {
     }
 }
 
+/// Specifies the number of results returned by a search
+#[derive(Debug, Copy, Clone)]
+pub enum Limit {
+    /// Always return all results
+    All,
+
+    /// Return up to the specified number of results
+    Max(i64),
+}
+
+impl Limit {
+    #[inline]
+    fn to_value(self) -> CFType {
+        match self {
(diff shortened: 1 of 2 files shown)


kaisq:
The test failure seems unrelated to my change. Is that a known issue?

security-framework/src/item.rs
```diff
@@ -132,7 +153,7 @@ impl ItemSearchOptions {
     ///
     /// If this is not called, the default limit is 1.
     #[inline(always)]
-    pub fn limit(&mut self, limit: i64) -> &mut Self {
```
kornelski, line 135: This is a breaking change. Maybe you could change it to `impl Into<Limit>` and implement `From<i64> for Limit`?

kaisq: Oh, that's a great idea! Added. It should be a non-breaking change now.

kornelski (approved):

PR #130, opened 2021-04-28
changed success return to piped file in execute_with_privileges function

x13a:
`Authorization.execute_with_privileges` function have to return piped file instead of `()`. With a file returned we can control execution flow like reading output or waiting for process exit.


kornelski:
Thank!

PR #131, opened 2021-05-13
Expose SecKey.h functionality on macOS

brandonweeks:
SecKey.h functionality has been supported on macOS since 10.12.

https://opensource.apple.com/source/Security/Security-58286.41.2/keychain/SecKey.h.auto.html


kornelski:
The change to `cfg` is bad — disabled iOS completely, and enabled it on **all** old versions of macOS

PR #132, opened 2021-06-04
Add initial support for code signing services

steven-joruk:
I'm not sure if the GuestAttributes is the right approach, maybe I'm just excessively averse to having to manually create CFDictionaries.

I'm intending to use this in another project to securely validate XPC client connections using their audit token.

### DIFF
--- a/security-framework/src/os/macos/code_signing.rs
+++ b/security-framework/src/os/macos/code_signing.rs
@@ -12,17 +12,17 @@ use core_foundation::{
 };
 use libc::pid_t;
 use security_framework_sys::code_signing::{
-    kSecCSAllowNetworkAccess, kSecCSBasicValidateOnly, kSecCSCheckAllArchitectures,
-    kSecCSCheckGatekeeperArchitectures, kSecCSCheckNestedCode, kSecCSCheckTrustedAnchors,
-    kSecCSConsiderExpiration, kSecCSDoNotValidateExecutable, kSecCSDoNotValidateResources,
-    kSecCSEnforceRevocationChecks, kSecCSFullReport, kSecCSNoNetworkAccess, kSecCSQuickCheck,
-    kSecCSReportProgress, kSecCSRestrictSidebandData, kSecCSRestrictSymlinks,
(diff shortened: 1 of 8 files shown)


security-framework/src/os/macos/code_signing.rs
```diff
@@ -0,0 +1,276 @@
...
+    SecCode, SecCodeRef
+}
+impl_TCFType!(SecCode, SecCodeRef, SecCodeGetTypeID);
+
+impl SecCode {
+    /// Retrieves the code object for the code making the call.
+    pub fn for_self(flags: Flags) -> Self {
+        let mut code = MaybeUninit::uninit();
+
+        let status = unsafe { SecCodeCopySelf(flags.bits(), code.as_mut_ptr()) };
```
steven-joruk, line 118: Woops, use cvt and Result

security-framework/src/os/macos/code_signing.rs
```diff
@@ -0,0 +1,269 @@
...
+    // - architecture
+    // - canonical
+    // - dynamic code
+    // - dynamic code info plist
+    // - hash
+    // - mach port
+    // - sub-architecture
+
+    /// The guest's audit token.
+    pub fn audit_token(&mut self, token: CFDataRef) {
```
kornelski, line 65: I think they should be named `set_*`. Other than that it seems fine. 

Maybe add `set_other<V: ToVoid<V>>(CFStringRef, V)` to allow adding other attributes.

steven-joruk:
I'm not sure what's going on with the stable test failures. I tried using [ctest2](https://github.com/JohnTitor/ctest2) instead which didn't help.

Using `super::` instead of `base::` works around some errors, but it'd require a lot of changes and doesn't solve the underlying issue. I'll keep looking.

steven-joruk:
The lints are failing on rust 1.38 due to a dependency using an unsupported clippy config. Is it okay to bump the rust version for the lints job? Is 1.42 okay?

kornelski:
Yeah, bumping clippy should be fine. Thank you for the PR.

PR #133, opened 2021-06-05
Fix creating GuestAttributes

steven-joruk:
My last PR had a mistake - there's no way for users to create a GuestAttributes instance, sorry about that.

I'm just testing this now, but I wanted to open it before you published a new version. I'll update the title when I'm done later today.

### DIFF
--- a/security-framework/src/os/macos/code_signing.rs
+++ b/security-framework/src/os/macos/code_signing.rs
@@ -124,10 +124,18 @@ impl GuestAttributes {
     // - mach port
     // - sub-architecture
 
+    /// Creates a new, empty `GuestAttributes`. You must add values to it in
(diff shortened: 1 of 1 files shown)


steven-joruk:
I've tested this locally now using https://github.com/dfrankland/xpc-connection-rs/pull/32

Unrelated, but I was caught out by this kind of pattern while using `GuestAttribute::set_audit_token()`, and I have been in the past as well:
```rust
// Create a dangling reference - it could be any concrete type, but just CFData.
let data_ref = CFData::from_buffer(...).as_concrete_TypeRef();

// Crashes when it tries to use the data_ref
some_safe_func(data_ref);
```

kornelski:
Can you add tests for it?

PR #134, opened 2021-06-06
Sec code tests

steven-joruk:
Add more tests for SecCode and add a Debug impl so that unwrap_err can be used.


kornelski:
Thank you

PR #135, opened 2021-06-06
Compile for Apple Silicon during CI builds

steven-joruk:
I considered combining it with the iOS job as a matrix, but I suppose the Apple Silicon target will become the main target in a few years.


kornelski (approved):

kornelski:
Thanks

PR #139, opened 2021-08-18
Add subject, issuer, and serial_number methods

brandonweeks:
These methods expose the subject, issuer and serial number fields in
their DER-encoded form.

Happy to write tests, just not sure what style would fit best with the package. Would you prefer comparing against a hard coded values (byte vectors or files) or adding a test dependency on a package that can programmatically build X.509 names?

### DIFF
--- a/security-framework-sys/src/certificate.rs
+++ b/security-framework-sys/src/certificate.rs
@@ -1,10 +1,10 @@
-#[cfg(target_os = "macos")]
+#[cfg(any(target_os = "macos", target_os = "ios"))]
 use core_foundation_sys::array::CFArrayRef;
 use core_foundation_sys::base::{CFAllocatorRef, CFTypeID, OSStatus};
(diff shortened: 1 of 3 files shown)


kornelski:
Thanks for the PR.

Dev dependencies are fine. I'm happy with any tests as long as it doesn't depend on certificates that would expire.

brandonweeks:
Added tests. The `x509_parser` crate doesn't do any verification, so there shouldn't be any issues with expiration.

kornelski:
Thank you

djc:
This PR caused all downstream users to pull in 4 extra crates (num, num-bigint, num-iter, num-rational) in order to support the new `SecCertificate::serial_number()` method. I don't think that's a trade-off I would make -- maybe reconsider?

kornelski:
Ok, I've marked it as deprecated.

djc:
Thanks for the quick response!

PR #142, opened 2021-08-23
Add fingerprint method to SecCertificateExt

brandonweeks:
Add a method for returning the SHA-256 fingerprint of a certificate on macOS. If preferable, this could be implemented in a cross-platform way by taking a dependency on CommonCrypto and calling `CC_SHA256` directly.

Verified with:
```shell
% openssl x509 -in security-framework/test/server.der -inform der -noout -fingerprint -sha256
SHA256 Fingerprint=AF:9D:D1:80:A3:26:AE:08:B3:7E:63:98:F9:26:2F:8B:9D:4C:55:67:4A:23:3A:7C:84:97:50:24:F8:73:65:5D
```

### DIFF
--- a/security-framework/src/os/macos/certificate.rs
+++ b/security-framework/src/os/macos/certificate.rs
@@ -3,6 +3,7 @@
 use core_foundation::array::{CFArray, CFArrayIterator};
 use core_foundation::base::TCFType;
 use core_foundation::base::ToVoid;
+use core_foundation::data::CFData;
 use core_foundation::dictionary::CFDictionary;
 use core_foundation::error::CFError;
 use core_foundation::string::CFString;
@@ -15,6 +16,7 @@ use crate::certificate::SecCertificate;
 use crate::cvt;
 use crate::key::SecKey;
 use crate::os::macos::certificate_oids::CertificateOid;
+use crate::os::macos::digest_transform::{Builder, DigestType};
 
 /// An extension trait adding OSX specific functionality to `SecCertificate`.
 pub trait SecCertificateExt {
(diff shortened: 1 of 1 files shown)


security-framework/src/os/macos/certificate.rs
```diff
@@ -30,6 +32,9 @@ pub trait SecCertificateExt {
     /// subset.
     fn properties(&self, keys: Option<&[CertificateOid]>)
         -> Result<CertificateProperties, CFError>;
+
+    /// Returns the SHA-256 fingerprint of the certificate.
+    fn fingerprint(&self) -> Result<String, CFError>;
```
kornelski, line 37: I think it'd be fine to return raw bytes, and leave hex-encoding to callers.

security-framework/src/os/macos/certificate.rs
```diff
@@ -83,6 +88,15 @@ impl SecCertificateExt for SecCertificate {
             }
         }
     }
+
+    /// Returns the SHA-256 fingerprint of the certificate.
+    fn fingerprint(&self) -> Result<String, CFError> {
+        let data = CFData::from_buffer("The quick brown fox jumps over the lazy dog".as_bytes());
+        let hash = Builder::new()
+            .type_(DigestType::md5())
```
kornelski, line 96: md5?

kornelski (changes requested):

security-framework/src/os/macos/certificate.rs
brandonweeks: Forgot to stage this file to the final commit. 😅 Thanks.

security-framework/src/os/macos/certificate.rs
brandonweeks: Done.

PR #145, opened 2021-08-24
Set SSLCipherSuite width to u16 on macOS AArch64

brandonweeks:
Fixes:
- os::macos::secure_transport::test::negotiated_cipher
- secure_transport::test::test_builder_blacklist_ciphers

on macOS AArch64.


kornelski:
Thanks

PR #149, opened 2021-08-24
Fix conditional compilation with OSX_10_12

brandonweeks:
This resolves some mistakes I made in #139 that only are apparent when the `OSX_10_12` feature is selected.


kornelski:
Thank you

PR #152, opened 2021-09-10
Implement `Hash` for `SecCertificate`, `SecIdentity` and `SecKey`

### LINKED ISSUE #151: impl common traits for public facing types
I'm working on implementing [common traits](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits) for public facing types via [this](https://github.com/sfackler/rust-native-tls/issues/207) issue on `rust-native-tls`.

Unfortunately this doesn't seem trivial, since some types are declared using the `declare_TCFType` macro, so deriving them is not possible. I tried implementing them by hand, but this doesn't seem to have an effect.

<details>
<summary>Implementation of traits for `SecIdentity`</summary>

```rs
declare_TCFType! {
    /// A type representing an identity.
    ///
    /// Identities are a certificate paired with the corresponding private key.
    SecIdentity, SecIdentityRef
}
impl_TCFType!(SecIdentity, SecIdentityRef, SecIdentityGetTypeID);

impl Copy for SecIdentity {}
impl Clone for SecIdentity {}
impl Eq for SecIdentity {}
impl PartialEq for SecIdentity {}
impl Ord for SecIdentity {}
impl PartialOrd for SecIdentity {}
impl Hash for SecIdentity {}
impl Debug for SecIdentity {}
impl Default for SecIdentity {}
```
</details>

<details>
<summary>Error describing that `SecIdentity` doesn't implement `Copy`</summary>

```
error[E0204]: the trait `Copy` may not be implemented for this type
  --> src/imp/security_framework.rs:78:10
   |
78 | #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
   |          ^^^^
79 | pub struct Identity {
80 |     identity: SecIdentity,
   |     --------------------- this field does not implement `Copy`
81 |     chain: Vec<SecCertificate>,
   |     -------------------------- this field does not implement `Copy`
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
</details>

I must say that I'm not familiar with this macro, so it could just be that I'm missing something. Do you know a way around this issue?

Thanks for feedback of any kind!

garritfra:
closes #151 

I'm not quite sure how to implement this trait for `SecKey`. Any hints are greatly appreciated!

### DIFF
--- a/security-framework/src/certificate.rs
+++ b/security-framework/src/certificate.rs
@@ -8,6 +8,7 @@ use core_foundation_sys::base::kCFAllocatorDefault;
 use security_framework_sys::base::{errSecParam, SecCertificateRef};
 use security_framework_sys::certificate::*;
 use std::fmt;
+use std::hash;
 use std::ptr;
 
 use crate::base::{Error, Result};
@@ -53,6 +54,12 @@ impl fmt::Debug for SecCertificate {
         fmt.debug_struct("SecCertificate")
             .field("subject", &self.subject_summary())
             .finish()
+    }
+}
+
+impl hash::Hash for SecCertificate {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.to_der().hash(state);
     }
 }
 
--- a/security-framework/src/identity.rs
+++ b/security-framework/src/identity.rs
@@ -4,6 +4,7 @@ use core_foundation::base::TCFType;
 use security_framework_sys::base::SecIdentityRef;
 use security_framework_sys::identity::*;
 use std::fmt;
+use std::hash;
 use std::ptr;
 
 use crate::base::Result;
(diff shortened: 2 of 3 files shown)


kornelski:
I'm unsure about this. 

The hash implementation calls system methods to extract data from these objects. Does Apple guarantee that these are deterministic? (e.g. no random nonce inserted somewhere?).

Also this could be relatively expensive (e.g. building a DER representation). Hashing is usually meant to be on the fast path.

I don't think these objects are appropriate for being hashed.

garritfra:
Maybe I'm stuck in tunnel-vision.
The reason for this proposal is this PR: https://github.com/sfackler/rust-native-tls/pull/208

My specific use-case is user authentication based on client certificates. As for user ids, I thought hashing the cert would do the trick. Do you know any other way to accomplish this?

garritfra:
Okay, I think I'm completely on a wrong track. I'll have to dig into TLS to better understand what's going on. Sorry for the inconvenience!

PR #155, opened 2021-12-23
iOS password support

### LINKED ISSUE #70: Keychain Services support
FYI, I've started writing a crate which provides a somewhat idiomatic Keychain Services binding:

https://github.com/iqlusioninc/keychain-services-rs

I'd be interested in trying to contribute its functionality upstream to this crate, or otherwise finding a way to share code (e.g. `SecKey`).

See also: https://github.com/iqlusioninc/keychain-services-rs/issues/3

brotskydotcom:
This PR adds support for setting/getting generic and internet passwords on iOS (and macOS) without use of the macOS file-based (legacy) keychain APIs.  It does most of the work to fix #70, although it doesn't go all the way in terms of adding complete support for finding/returning all attributes of all items.

Because there's no way to actually test iOS support without using an iOS app, this PR also adds a Swift app with tests to the workspace tree, as well as a shim-library crate with a CoreFoundation interface that can be called directly from Swift.

It should be possible to enhance the GitHub CI action to do iOS as well as macOS testing, but this PR doesn't do that.


brotskydotcom:
@kornelski I believe this is ready for a first review.

security-framework/src/item.rs
```diff
@@ -175,7 +174,8 @@ impl ItemSearchOptions {
     /// Sets kSecAttrAccessGroup to kSecAttrAccessGroupToken
     #[inline(always)]
     pub fn access_group_token(&mut self) -> &mut Self {
-        self.access_group = unsafe { Some(CFString::wrap_under_get_rule(kSecAttrAccessGroupToken)) };
```
kornelski, line 178: Please don't change formatting of unrelated things

ios-shim/src/iospw.h
```diff
@@ -0,0 +1,14 @@
+//
+// Created by Daniel Brotsky on 12/18/21.
+//
+
+#ifndef RUST_SECURITY_FRAMEWORK_IOSPW_H
+#define RUST_SECURITY_FRAMEWORK_IOSPW_H
+
+#include <CoreFoundation/CoreFoundation.h>
+
+extern OSStatus SecSetGenericPassword(CFStringRef service, CFStringRef account, CFDataRef password);
```
kornelski, line 10: Can you add comments why this shim is needed?

ios-shim/src/lib.rs
```diff
@@ -0,0 +1,60 @@
+use core_foundation::base::{CFRetain, OSStatus, TCFType};
+use core_foundation::data::{CFData, CFDataRef};
+use core_foundation::string::{CFString, CFStringRef};
+use security_framework_sys::base::{errSecBadReq, errSecSuccess};
+
+#[no_mangle]
+pub extern "C" fn SecSetGenericPassword(
```
kornelski, line 7: This uses Apple's `Sec*` naming scheme. All symbols in C are global, so there's a risk of a collision with Apple's functions. If this function is just for Rust, please add a custom name prefix, e.g. `rust_sec_framework_set_generic_password`

security-framework/src/item.rs
brotskydotcom: Sorry!  I use `rustfmt` and it made some changes which I mistakenly assumed were expected.  I will revert any changes it made in older code.

ios-shim/src/iospw.h
brotskydotcom: Yep, will do.

ios-shim/src/lib.rs
brotskydotcom: This function is actually just for use from Swift and Objective C/C++ code, both of which use camel case naming.  I am happy not to use Apple's naming convention, but would prefer to keep using camel case.  How about `Rust` or `RustSec` or `RustShim` as a prefix, as in `RustShimSetGenericPassword`?  Do you have a preference?

kornelski: RustShim is fine.

security-framework/src/passwords.rs
```diff
@@ -0,0 +1,351 @@
...
+    if security_domain.is_some() {
+        query.push((
+            unsafe { CFString::wrap_under_get_rule(kSecAttrSecurityDomain) },
+            CFString::from(security_domain.unwrap()).as_CFType(),
+        ))
+    }
+    if port.is_some() {
+        query.push((
+            unsafe { CFString::wrap_under_get_rule(kSecAttrPort) },
+            CFNumber::from(port.unwrap() as i32).as_CFType(),
```
kornelski, line 198: Please use `if let` to avoid unwrap.

security-framework/src/item.rs
brotskydotcom: Reverted in my latest commits.

ios-shim/src/lib.rs
brotskydotcom: OK, will use `RustShim`

security-framework/src/passwords.rs
brotskydotcom: Great catch, thanks.

brotskydotcom:
@kornelski I've addressed your review comments, added docs, and fixed the `clippy` issues that were breaking the build.  I think this is ready to go.  Let me know if you want me to enhance the CI script to run the iOS tests via XCode.

security-framework/src/passwords.rs
```diff
@@ -0,0 +1,361 @@
...
+/// keychain entry exists, fails with error code `errSecItemNotFound`.
+pub fn get_generic_password(service: &str, account: &str) -> Result<Vec<u8>> {
+    let mut query = generic_password_query(service, account);
+    query.push((
+        unsafe { CFString::wrap_under_get_rule(kSecReturnData) },
+        CFBoolean::from(true).as_CFType(),
+    ));
+    let params = CFDictionary::from_CFType_pairs(&query);
+    let mut ret: CFTypeRef = std::ptr::null();
+    cvt(unsafe { SecItemCopyMatching(params.as_concrete_TypeRef(), &mut ret) })?;
```
kornelski, line 44: `ret` needs to be retained using `wrap_under_create_rule`, but `get_password_internal` doesn't do that.

kornelski (changes requested):

security-framework/src/passwords.rs
```diff
@@ -0,0 +1,361 @@
...
+        protocol,
+        authentication_type,
+    );
+    query.push((
+        unsafe { CFString::wrap_under_get_rule(kSecReturnData) },
+        CFBoolean::from(true).as_CFType(),
+    ));
+    let params = CFDictionary::from_CFType_pairs(&query);
+    let mut ret: CFTypeRef = std::ptr::null();
+    cvt(unsafe { SecItemCopyMatching(params.as_concrete_TypeRef(), &mut ret) })?;
```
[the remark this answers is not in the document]
brotskydotcom: Great catch, thanks!  Resolved in latest commit, which adds a number of safety checks to `get_password_and_release` (renamed from `get_password_internal`).

security-framework/src/passwords.rs
brotskydotcom: Yep, same as above.

brotskydotcom:
@kornelski While working on this PR, there were some clippy warnings in older code.  I'll open a separate issue with those.

kornelski (approved):

kornelski:
Thank you

PR #156, opened 2022-01-15
Expand Trust and Policy API bindings

complexspaces:
This PR adds bindings to both `security-framework-sys` and `security-framework` for most* of the rest of the `SecTrust` related and policy creation APIs.

The only thing I'm not sure about is the correct way to expose methods like `SecTrustSetSignedCertificateTimestamps` and `SecTrustSetOCSPResponse`, which accept arrays of arbitrary data blobs, in `security-framework`'s public API. Other methods like this have corresponding `core-foundation` types, like `SecCertificate`. Given this, for now I've opted to expose `CFData` in the public API methods to allow the caller to control all allocation. Otherwise, I believe the methods would need to take an iterator and allocate a `Vec<CFData>` internally to work correctly.

Thanks!

### DIFF
--- a/security-framework/src/trust.rs
+++ b/security-framework/src/trust.rs
@@ -4,6 +4,9 @@ use core_foundation::array::CFArray;
 #[cfg(target_os = "macos")]
 use core_foundation::array::CFArrayRef;
 use core_foundation::base::TCFType;
+#[cfg(any(feature = "OSX_10_9", target_os = "ios"))]
+use core_foundation::data::CFData;
+use core_foundation::date::CFDate;
 use core_foundation_sys::base::{Boolean, CFIndex};
 
 use security_framework_sys::trust::*;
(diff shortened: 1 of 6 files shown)


security-framework/src/trust.rs
```diff
@@ -4,6 +4,9 @@ use core_foundation::array::CFArray;
 #[cfg(target_os = "macos")]
 use core_foundation::array::CFArrayRef;
 use core_foundation::base::TCFType;
+#[cfg(any(feature = "OSX_10_9", target_os = "ios"))]
+use core_foundation::data::CFData;
+use core_foundation::date::CFDate;
```
kornelski, line 9: dupe import

kornelski:
I'm not sure about CFData. I'm not very worried about allocation, because it's probably negligible compared to all the other work these methods are doing.

So I wonder more whether this is convenient for users. Where do these timestamps come from? If there's an API that returns CFData? Or are users expected to use a Rust crate for it? If they're created on the Rust side, it'd be nice to take `impl Iterator<Item=impl AsRef<[u8]>>`.

complexspaces:
There is not a specific Apple API that these come from, they can come from anywhere a TLS handshake is done, be that with the OS or a crate like `rustls` which provides them as an iterator of slices. I agree iterators are the best for usability, I was just unsure about your stance on callers vs library responsibility for handling allocation.

I've pushed up a change that switches both methods that accept a set of byte blobs to use `Iterator<Item = impl AsRef<[u8]>>>`.  I don't necessarily see how `AsRef<[u8]>` makes things much more ergonomic, though, but it may just be me. It seems most callers will already have `Vec<u8>` or `&[u8]` on hand already, so slice coercion would take care of the rest constructing an iterator. That being said, `AsRef<[u8]>` looks nicer in the type signature than a slice lifetime.

PR #158, opened 2022-02-04
Fix iOS support for updating of generic passwords

### LINKED ISSUE #157: iOS password support fails on update of generic password
While doing iOS-based testing of my contributed iOS-support, I found a bug that causes failures to update the value of an existing generic password.  This only happens with generic passwords, not internet passwords.  I am contributing a fix immediately.

brotskydotcom:
This fixes #157 by removing the spurious `kSecReturnData` attribute in queries for generic passwords; this attribute was accepted in update queries on macOS but not on iOS.

This also streamlines and improves iOS-based test coverage for the security framework, by providing an iOS test harness similar to the Rust test harness.  For now, the only tests run on iOS are for the generic and internet password support, but the framework allows testing of all iOS functionality, should that be needed in future.  (Another way of saying this is that the `iostest` crate is now analogous to the `systest` crate but for iOS.)


brotskydotcom:
@kornelski, this is now ready for your review.  The functional change is very small, but there are lots of improvements in the iOS test harness, and I've tightened the error-code checks in the underlying password code.

kornelski:
Thank you for the fixes and tests.

PR #160, opened 2022-07-12
SMJobBless

hankbao:
Since we cannot do `as_concrete_TypeRef()` on `Authorization`, I guess implementing `SMJobBless` directly within it would be okay.

### DIFF
--- a/security-framework/Cargo.toml
+++ b/security-framework/Cargo.toml
@@ -21,6 +21,7 @@ bitflags = "1.3.2"
 libc = "0.2.100"
 log = { version = "0.4.14", optional = true }
 num-bigint = { version = "0.4.3", optional = true }
+service-management-sys = { version = "0.1", optional = true }
 
 [dev-dependencies]
 tempdir = "0.3.7"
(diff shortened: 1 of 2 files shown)


kornelski:
Thanks for the PR. 


Why the string type changes? Taking both str and CStringRef is odd.

kornelski:
Also it'd be nice to return the boolean from smjobbbless.

hankbao:
> Why the string type changes? Taking both str and CStringRef is odd.

That's because `domain` could only be `kSMDomainSystemLaunchd`, a `CFStringRef` exported by service-management-sys (from ServiceManagement.framework). So taking `domain` as a `CStringRef` would be more straightforward here.

> Also it'd be nice to return the boolean from smjobbbless.

If `SMJobBless` fail, an error from `CFError` will be returned. Same behavior as the `SecTransformSetAttribute` call in `SecTransform:: set_attribute`.

kornelski:
Is `kSMDomainSystemLaunchd` literally the only useful value, or is `kSMDomainUserLaunchd` usable too? If only one, then you could just hardcode it inside the function. If both, you could make a Rust enum for them. It would be nice not to use ObjC types in a higher-level Rust API, especially that it's not just any string, but a specific key, so it should offer more type-safety than ObjC does.

As for returning a bool, I mean returning `Result<bool, …>`

kornelski:
Oh, I see. I've misinterpreted the docs and thought the `true` worked like `hashmap.insert` result.

kornelski:
service-management-sys doesn't compile on iOS

kornelski:
Right. Fixed 438db710bcadff9327bad2b3aa3ba68d427b2d0d

PR #161, opened 2022-11-18
Add key generation

kcking:
move KeyType to crate::key since it is supported on iOS

### DIFF
--- a/security-framework/src/key.rs
+++ b/security-framework/src/key.rs
@@ -13,10 +13,10 @@ use core_foundation::dictionary::CFDictionary;
 #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
 use core_foundation::error::{CFError, CFErrorRef};
 
-use security_framework_sys::item::{
+use security_framework_sys::{item::{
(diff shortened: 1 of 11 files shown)


kornelski:
Looks good!

security-framework/src/key.rs
```diff
@@ -28,7 +94,58 @@ impl_TCFType!(SecKey, SecKeyRef, SecKeyGetTypeID);
 unsafe impl Sync for SecKey {}
 unsafe impl Send for SecKey {}
 
+/// Which Keychain to store the generated key in. On iOS this struct is empty as
+/// there is only one keychain.
+pub struct StoreInKeychain {
+    #[cfg(target_os="macos")]
```
kornelski, line 100: maybe `generate` could rely on taking `None` instead of empty struct?

kcking:
Thanks for taking a look! I'm still iterating on the implementation a bit as I use it in a project. Hoping to support secure enclave keys as well :)

kcking:
I am breaking this long chain of commits down into multiple PRs, the first of which is https://github.com/kornelski/rust-security-framework/pull/164

PR #163, opened 2022-11-20
Make Error::inner_message universal

### LINKED ISSUE #162: Why isn't SecCopyErrorMessageString called for non MacOS targets?
Hello,

`SecCopyErrorMessageString` was first introduced [7 years ago](https://github.com/kornelski/rust-security-framework/commit/eeaea787308a24caa4005a1c4be23ec0c80504f6) to make error display more "reasonable".

However, since then, it's still only limited to MacOS 
https://github.com/kornelski/rust-security-framework/blob/0f51c36018be67dbfcbc4789b460767c04bc217f/security-framework/src/base.rs#L46-L67

as it returns None for other targets.
Is there a reason for this? In case you think it's interesting, here's [a PR](https://github.com/kornelski/rust-security-framework/pull/163) with my suggestion.

Regards,
Nabil

n-eq:
Closes #162

### DIFF
--- a/security-framework/src/base.rs
+++ b/security-framework/src/base.rs
@@ -43,7 +43,7 @@ impl Error {
         self.inner_message()
     }
 
-    #[cfg(target_os = "macos")]
+    #[inline(always)]
     #[cold]
     fn inner_message(self) -> Option<String> {
         use core_foundation::base::TCFType;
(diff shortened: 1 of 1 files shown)


security-framework/src/base.rs
```diff
@@ -43,7 +43,7 @@ impl Error {
         self.inner_message()
     }
 
-    #[cfg(target_os = "macos")]
+    #[inline(always)]
```
kornelski, line 46: Please don't add inline here

n-eq, line 46:
```suggestion
```

n-eq: Sure, done

n-eq:
I fixed a red test, can you please have a look @kornelski?

PR #166, opened 2022-11-24
Add SecKey::application_label method

brandonweeks:
https://developer.apple.com/documentation/security/ksecattrapplicationlabel

The `kSecAttrApplicationLabel` attribute is used extensively in dealing with Keychain identities as it is the attribute used to pair a `SecKey` and `SecCertificate` into a `SecIdentity`. This [Apple Developer Forums post](https://developer.apple.com/forums/thread/69642?answerId=202264022#202264022) documents this behavior.

### DIFF
--- a/security-framework/src/key.rs
+++ b/security-framework/src/key.rs
@@ -32,6 +32,8 @@ use security_framework_sys::key::{
     SecKeyCreateSignature, Algorithm, SecKeyCreateRandomKey,
     SecKeyCopyPublicKey,
 };
+#[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
+use security_framework_sys::item::kSecAttrApplicationLabel;
 use std::fmt;
 
 use crate::{base::Error, item::Location};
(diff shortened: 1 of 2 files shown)


security-framework/src/key.rs
```diff
@@ -119,6 +121,19 @@ impl SecKey {
...
 
+    /// Returns the programmatic identifier for the key. For keys of class 
+    /// kSecAttrKeyClassPublic and kSecAttrKeyClassPrivate, the value is the 
+    /// hash of the public key.
+    #[cfg(any(feature = "OSX_10_12", target_os = "ios"))]
+    pub fn application_label(&self) -> Vec<u8> {
+        let attributes = self.attributes();
+        unsafe {
+            CFData::wrap_under_get_rule(
+                attributes.get(kSecAttrApplicationLabel.to_void()).cast()
```
kornelski, line 132: Is the attribute guaranteed to exist? Could get return nil?

brandonweeks:
```swift
var params: [String: Any] = [
    kSecAttrKeyType as String: kSecAttrKeyTypeAES,
    kSecAttrKeySizeInBits as String: 256,
    kSecAttrIsPermanent as String: false,
]

var error: Unmanaged<CFError>?
guard let key = SecKeyGenerateSymmetric(params as CFDictionary, &error) else {
    throw error!.takeRetainedValue() as Error
}
let attrs = SecKeyCopyAttributes(key)! as NSDictionary
attrs[kSecAttrApplicationLabel]!
```

The attribute is guaranteed to exist for "permanent" keys, however it does not appear to exist non-permanent keys as demonstrated by the above code snippet.

I've changed the return type to be `Option<Vec<u8>>` and used `.find` instead of `.get` on the attributes dictonary.

PR #167, opened 2022-11-24
Add add_item wrapper around SecItemAdd

kcking:
Move Location struct from key.rs to item.rs and re-use it for this SecItemAdd abstraction.

Create an `ItemAddOption`s builder struct.


kornelski:
Thanks!

PR #178, opened 2023-03-06
Draft: Support access control options

Mikescops:
Hello,

I'm trying to implement the support of the Keychain Item Accessibility ([see Apple documentation here](https://developer.apple.com/documentation/security/keychain_services/keychain_items/restricting_keychain_item_accessibility?language=objc)).

The goal is for instance to ask for user presence (faceId, touchId) when requesting a password from the keychain. This options are widely supported on ios/macos/tvos/... so I thought it makes sense to include it in this library.

Disclaimer: I'm not familiar with Rust at all, so I did my best to get into it, here is a draft so let me know if I'm heading in the right direction and any guidance/help is welcomed!

I didn't added tests yet, didn't want to spend too much time if I'm already not doing things right

### DIFF
--- a/security-framework-sys/src/access_control.rs
+++ b/security-framework-sys/src/access_control.rs
@@ -1,4 +1,4 @@
-use core_foundation_sys::base::{CFAllocatorRef, CFTypeRef};
+use core_foundation_sys::base::{CFAllocatorRef, CFTypeRef, CFTypeID};
 use core_foundation_sys::error::CFErrorRef;
 use core_foundation_sys::base::CFOptionFlags;
 
@@ -21,5 +21,12 @@ mod access_control_flags {
 pub use access_control_flags::*;
 
 extern "C" {
-    pub fn SecAccessControlCreateWithFlags(allocator: CFAllocatorRef, protection: CFTypeRef, flags: CFOptionFlags, error: *mut CFErrorRef) -> SecAccessControlRef;
+    pub fn SecAccessControlGetTypeID() -> CFTypeID;
+
+    pub fn SecAccessControlCreateWithFlags(
+        allocator: CFAllocatorRef,
+        protection: CFTypeRef,
+        flags: CFOptionFlags,
+        error: *mut CFErrorRef
+    ) -> SecAccessControlRef;
 }
--- a/security-framework-sys/src/access_control.rs
+++ b/security-framework-sys/src/access_control.rs
@@ -1,0 +1,25 @@
+use core_foundation_sys::base::{CFAllocatorRef, CFTypeRef};
+use core_foundation_sys::error::CFErrorRef;
+use core_foundation_sys::base::CFOptionFlags;
+
+use crate::base::SecAccessControlRef;
+
+mod access_control_flags {
+    use super::CFOptionFlags;
+
+    pub const kSecAccessControlUserPresence: CFOptionFlags = 1 << 0;
+    pub const kSecAccessControlBiometryAny: CFOptionFlags = 1 << 1;
+    pub const kSecAccessControlBiometryCurrentSet: CFOptionFlags = 1 << 3;
+    pub const kSecAccessControlDevicePasscode: CFOptionFlags = 1 << 4;
+    pub const kSecAccessControlWatch: CFOptionFlags = 1 << 5;
+    pub const kSecAccessControlOr: CFOptionFlags = 1 << 14;
+    pub const kSecAccessControlAnd: CFOptionFlags = 1 << 15;
+    pub const kSecAccessControlPrivateKeyUsage: CFOptionFlags = 1 << 30;
+    pub const kSecAccessControlApplicationPassword: CFOptionFlags = 1 << 31;
+}
+
+pub use access_control_flags::*;
+
+extern "C" {
+    pub fn SecAccessControlCreateWithFlags(allocator: CFAllocatorRef, protection: CFTypeRef, flags: CFOptionFlags, error: *mut CFErrorRef) -> SecAccessControlRef;
+}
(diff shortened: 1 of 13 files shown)


kornelski:
Thanks for the PR.

I'm happy to add more of the security framework functionality, but I'd prefer not to make breaking changes to existing Rust APIs. If you need to modify existing tests to compile the code, that's a breaking change.

security-framework-sys/src/access_control.rs
```diff
@@ -0,0 +1,25 @@
+use core_foundation_sys::base::{CFAllocatorRef, CFTypeRef};
+use core_foundation_sys::error::CFErrorRef;
+use core_foundation_sys::base::CFOptionFlags;
+
+use crate::base::SecAccessControlRef;
+
+mod access_control_flags {
```
kornelski, line 7: if you have `pub use` right after it, then the module seems redundant.

security-framework-sys/src/base.rs
```diff
@@ -32,6 +32,8 @@ pub type SecCertificateRef = *mut OpaqueSecCertificateRef;
 pub enum OpaqueSecAccessRef {}
 pub type SecAccessRef = *mut OpaqueSecAccessRef;
 
+pub type SecAccessControlRef = CFStringRef;
```
kornelski, line 35: in Apple's docs this type is not a string. Note that in Rust `type` is just a weak alias, like C's typedef, and offers no type safety.

security-framework/src/passwords.rs
```diff
@@ -26,9 +29,9 @@ use security_framework_sys::keychain_item::{
 
 /// Set a generic password for the given service and account.
 /// Creates or updates a keychain entry.
-pub fn set_generic_password(service: &str, account: &str, password: &[u8]) -> Result<()> {
+pub fn set_generic_password(service: &str, account: &str, password: &[u8], access_control_options: Option<AccessControlOptions>) -> Result<()> {
```
kornelski, line 32: You'll need to add this as a separate method, because `set_generic_password` is a stable API (promised to never change).

Perhaps it'd be a good time to introduce a builder-like pattern here, similar to what Apple does with the `query`. Create a `PasswordOptions` type that wraps the `Vec<(CFString, CFType)>` and sets query properties individually, and then a method like `set_password(PasswordOptions)`. This will be extensible for future properties that Apple may add.

security-framework/src/passwords.rs
```diff
@@ -204,14 +208,47 @@ fn internet_password_query(
...
-fn set_password_internal(query: &mut Vec<(CFString, CFType)>, password: &[u8]) -> Result<()> {
+fn set_password_internal(query: &mut Vec<(CFString, CFType)>, password: &[u8], access_control_options: Option<AccessControlOptions>) -> Result<()> {
     let query_len = query.len();
     query.push((
         unsafe { CFString::wrap_under_get_rule(kSecValueData) },
         CFData::from_buffer(password).as_CFType(),
     ));
+
+    if let Some(options) = access_control_options {
+        let error = ptr::null_mut();
```
kornelski, line 245: this doesn't look right. If you're trying to capture the error, this should be defined in an outer scope and use `&mut error` later. Otherwise just use `ptr::null_mut()` without assigning it to a variable first.

kornelski (changes requested):

security-framework-sys/src/base.rs
Mikescops: but then the later object expects it to be a string in the attribute map, that's where I got confused (i initially used an opaque), the Apple documentation is not super explicit about how it gets converted

security-framework-sys/src/access_control.rs
Mikescops: oh ok thanks, I was reusing the syntax I saw in another file

Mikescops:
> If you need to modify existing tests to compile the code, that's a breaking change.

Totally agree, thanks for the tips!

security-framework/src/passwords.rs
Mikescops: Sure I'll try to do something like this

security-framework-sys/src/base.rs
Mikescops: ok found my way

security-framework-sys/src/access_control.rs
Mikescops: actually i'm not sure how to address this

security-framework/src/passwords.rs
Mikescops: so i tried my best for this one, learning on the go, see my recent changes

Mikescops:
@kornelski I tried to make your suggestions, I think I got something decent

It's not touching the function signatures, and the tests are passing, i restructured the password options (hope it's what you wanted).

Now I'm not sure exactly how to expose my new `set_access_control_options` method to the outside, looking for your advice

dmolokanov:
@Mikescops / @kornelski Do you guys have plans to merge this PR.

Mikescops:
@dmolokanov I'd love to do so yes, but i need feedback first

kornelski:
Looks good, thank you.

gibfahn:
>The goal is for instance to ask for user presence (faceId, touchId) when requesting a password from the keychain. This options are widely supported on ios/macos/tvos/... so I thought it makes sense to include it in this library.

Came here looking for the same thing, biometric support for keychain passwords.

Seems that this PR added PasswordOptions, but I can't find a way to actually plumb them in to the generic password creation API.

```rust
use security_framework::passwords_options::{AccessControlOptions, PasswordOptions};

fn main() {
    let access_control_options = AccessControlOptions::USER_PRESENCE;
    let mut entry = PasswordOptions::new_generic_password("my_service3", "test1");
    entry.set_access_control_options(access_control_options);
    // How to actually use this?
}
```

The above example also errors with:

`called Result::unwrap() on an Err value: Error { code: -50, message: "One or more parameters passed to a function were not valid." }`

which comes from `security-framework-2.9.2/src/passwords_options.rs:124:13`

brianlyn:
I'm running into the same error @gibfahn . Did you ever figure this out? I'm thinking that it might be related to missing entitlements that are required to use certain flags but I'm struggling to find documentation and I could just be misusing the API.

thenicekat:
Hello, did we make any progress here?

Issue #182, opened 2023-05-13
Infinite run on read_to_end with pkcs12


ragoso:
I'm coming from a reqwest implementation where I'm getting a "`peer cert is valid, or was ignored if verification disabled`" error

To validate if the error comes from the security-framework, I created the following test, but for some reason it runs infinitely.

```
#[test]
    fn load_page_with_pkcs12()
    {
        let pkcs12_der = include_bytes!("../test/my_cert.pfx");
        let ctxx = ServerBuilder::from_pkcs12(pkcs12_der, "1234").unwrap();

        let mut ctx = p!(SslContext::new(
            SslProtocolSide::CLIENT,
            SslConnectionType::STREAM
        ));

        ctx.set_certificate(&ctxx.identity, &ctxx.certs).unwrap();

        p!(ctx.set_peer_domain_name("hom1.nfe.fazenda.gov.br"));
        let stream = p!(TcpStream::connect("hom1.nfe.fazenda.gov.br:443"));
        let mut stream = p!(ctx.handshake(stream));
        p!(stream.write_all(b"GET /NFeDistribuicaoDFe/NFeDistribuicaoDFe.asmx HTTP/1.0\r\nHost: hom1.nfe.fazenda.gov.br\r\n\r\n"));
        p!(stream.flush());
        let mut buf = vec![];
        p!(stream.read_to_end(&mut buf));
        println!("{}", String::from_utf8_lossy(&buf));

    }
```

```
 Compiling security-framework v2.9.0 (/Users/matheusragoso/Projects/rust-security-framework/security-framework)
    Finished test [unoptimized + debuginfo] target(s) in 1.65s
     Running unittests src/lib.rs (target/debug/deps/security_framework-df6cd6b7a6b0fc69)

running 1 test
test secure_transport::test::load_page_with_pkcs12 has been running for over 60 seconds
```

What could be happening?

kornelski:
If I run `curl -v hom1.nfe.fazenda.gov.br:443` I get no reply either, so I think it's just the server not responding.

ragoso:
> If I run `curl -v hom1.nfe.fazenda.gov.br:443` I get no reply either, so I think it's just the server not responding.

Maybe it has something to do with the region.
```
➜  ~ curl -v hom1.nfe.fazenda.gov.br:443
*   Trying 189.9.67.82:443...
* Connected to hom1.nfe.fazenda.gov.br (189.9.67.82) port 443 (#0)
[...]

PR #183, opened 2023-05-16
handle non-fatal errors on SSLRead

ragoso:
From #182 

Based on curl's solution on [this issue and PR](https://github.com/curl/curl/issues/3932)


kornelski:
I'm worried about continuing on `errSSLWouldBlock`. You get this with non-blocking sockets, so next read will very likely return the same error, and since you're calling read recursively, this can blow up the stack.

ragoso:
> I'm worried about continuing on `errSSLWouldBlock`. You get this with non-blocking sockets, so next read will very likely return the same error, and since you're calling read recursively, this can blow up the stack (and if it didn't, it'd be busy-looping).

You're right, let's just stop at the root cause.

PR #189, opened 2023-12-04
Add SecKeyOperationType and SecKeyIsAlgorithmSupported

sheagcraig:
Greetings! I've got a rustls resolver implementation that uses the included code to ensure an algorithm is supported and thought I'd upstream these definitions to you.

I don't have a toy-example handy; let me know what it would take to demonstrate this code is solid and get it included.

Thanks for your project!


kornelski:
Thanks

PR #193, opened 2024-03-22
Allow creating a SecAccessControl with a protection

alex:


kornelski:
Could it use a more specific type than CFTypeRef?

alex:
I'm not sure :-) `CFTypeRef` is what the underlying C API uses. Do you have a thought as to how we might model it differently here?

kornelski:
I'm not familiar with that API. In general in the higher level wrapper I'd prefer to hide ObjC-isms and have type safety where possible.

What kind of objects are you passing there? Where do you get them from?

alex:
For my use case, I'm passing a `CFString`: `kSecAttrAccessibleWhenUnlockedThisDeviceOnly`

kornelski:
So that's an enum:

https://developer.apple.com/documentation/security/keychain_services/keychain_items/item_attribute_keys_and_values#1679100

alex:
Ok, now uses an enum.

kornelski:
Thanks

alex:
Thank you!

PR #194, opened 2024-03-23
Upgrade to bitflags 2

alex:


kornelski:
This may not be fully semver-compatible, and I'd rather not make major release for such detail.

alex:
I believe with the added derives, it should be compatible. Not sure how to prove that though

alex:
As a note, I believe this fixes a clippy warning. When it gets merged, I'll do a follow up to get clippy and teh rest of CI green.

TxMat:
Also fixes a cargo deny warning

image

TxMat:
Thanks !

PR #197, opened 2024-03-27
security-framework::item::ItemSearchOptions: support trusted_only

grahamc:
Extends the search options interface to support the kSecMatchTrustedOnly query option.

### DIFF
--- a/security-framework/src/item.rs
+++ b/security-framework/src/item.rs
@@ -327,7 +327,7 @@ impl ItemSearchOptions {
             if let Some(ref trusted_only) = self.trusted_only {
                 params.push((
                     CFString::wrap_under_get_rule(kSecMatchTrustedOnly),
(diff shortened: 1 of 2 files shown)


security-framework/src/item.rs
```diff
@@ -316,6 +324,13 @@ impl ItemSearchOptions {
                 ));
             }
 
+            if let Some(ref trusted_only) = self.trusted_only {
+                params.push((
+                    CFString::wrap_under_get_rule(kSecMatchTrustedOnly),
+                    if *trusted_only { CFBoolean::true_value().into_CFType() } else { CFBoolean::true_value().into_CFType() },
```
kornelski, line 330: Both branches are true

grahamc: Wouldn't want some sneaky untrusted certs getting out. Belt and suspenders? (...oops!) Fixup'd.

PR #200, opened 2024-03-30
Add more item search / add options, and access group search

grahamc:


security-framework/src/item.rs
```diff
@@ -592,11 +633,30 @@ impl ItemAddOptions {
                 },
             }
         }
-
+        let account_name = self.account_name.as_deref().map(CFString::from);
```
kornelski, line 636: I've just noticed that this and previous code are vulnerable to use after free. 

to_void gives a pointer to a temporary copy made by `CFString::from`. 

It should be giving a pointer to original longer lived CFString stored in self, or remove it from self and pass ownership, preventing Rust from decrementing refcount.

kornelski (changes requested):

security-framework/src/item.rs
grahamc: Ok! so more like `pub struct ItemSearchOptions {`, which stores the CFString directly in the struct? Can do.

grahamc: Hey so unfortunately I'm not real confident in my ability to do this correctly. I'm not real familiar with CF or unsafe Rust. I took a look into this, but never felt good enough about what I was trying.

kornelski: Sorry, false alarm! I've checked the implementation of `dict.add`, and it increases the reference counter. So the combo of {new string, give pointer, drop string} is valid.

grahamc: 🎉  thank you!

PR #201, opened 2024-04-09
Added visionos support

simlay:
Support for the Apple Vision Pro was added as a tier 3 target with a `targe_os` of `visionos` to rust in https://github.com/rust-lang/rust/pull/121419. Needs https://github.com/rust-lang/libc/pull/3568 to do proper build for the target `aarch64-apple-visionos-sim` or `aarch64-apple-visionos`.

This is basically the same changes as #190 but with another `target_os`.


kornelski:
Thank you

Issue #203, opened 2024-05-08
iOS Build Error: 'kSecMatchSubjectWholeString' symbol not found


gezihuzi:
Hello,

I am developing a universal SDK that depends on the `rust-security-framework` for using Keychain functionalities. After updating this dependency recently, I encountered a compilation issue on the iOS platform.

### Error Output

```sh
Undefined symbols for architecture arm64:
  "_kSecMatchSubjectWholeString", referenced from:
      security_framework::item::ItemSearchOptions::search::h115ebc3a4f9ba099 in libxxx-kit_static.a[arm64][1774](security_framework-3f9eb5a707bf396a.security_framework.a39d3ed3f5cf992f-cgu.08.rcgu.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Environment
- iOS version: 17.4.1
- Xcode version: 15.3
- Language: Swift + Rust
- rust-security-framework version: 2.11.0

### Documentation Reference
Apple Developer Documentation link that states [kSecMatchSubjectWholeString](https://developer.apple.com/documentation/security/ksecmatchsubjectwholestring)  is only available on macOS.

Thank you for looking into this issue. I appreciate any guidance or updates you can provide.

gezihuzi:
#204 Merged

gezihuzi:
This PR #204 may require a new release to take effect. May I ask if @kornelski has any plans to release a new version recently? I will keep this issue open until the problem is confirmed resolved.

gezihuzi:
Note: Use the `cargo update -p security-framework --precise 2.10.0` command to roll back to version v2.10.0 to ensure iOS can compile correctly.

brotskydotcom:
Hi @kornelski, you may remember me, I'm the person who contributed the iOS support for this crate a few years ago. I see that @sfackler has marked this crate as needing a maintainer, and I see that things on the iOS side have gotten into a bit of disrepair. I am, in fact, willing to help maintain this crate, as I am still very active on both macOS and iOS and use this crate often.  Feel free to ping me via email if this is of any interest to you.

kornelski:
@brotskydotcom Yes, please. Help is very welcome.

brotskydotcom:
OK, I'll try to review the outstanding issues and make some recommendations/submit some PRs over the next week or so. I think probably there are a few issues (such as this one and #206) that are probably worth releasing the fixes for in the near term.

PR #204, opened 2024-05-08
fix: Exclude macOS-only Keychain symbol for iOS compatibility

gezihuzi:
Inclusion of the `kSecMatchSubjectWholeString` symbol when compiling for iOS, which is not supported according to Apple's documentation, is prevented. Conditional compilation directives are added to ensure that builds targeted for iOS do not use macOS exclusive Keychain functionalities.

This adjustment resolves a compilation failure encountered following an update of the 'rust-security-framework' dependency, thereby enhancing the cross-platform compatibility of the SDK.


gezihuzi:
This PR resolves #203

pablogeek:
whe is this going to be released?

Issue #206, opened 2024-07-14
iOS only: security_framework::passwords::get_password_and_release crashes on empty password


brotskydotcom:
On Rust 1.79.0, [line 163 in passwords.rs](https://github.com/kornelski/rust-security-framework/blob/ea683f291bb8dff51d3239e556c3236e09274c16/security-framework/src/passwords.rs#L163C1-L164C1)

```rust
vec.extend_from_slice(val.bytes());
```
crashes if `val` is empty, with the message:

> unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`

This can easily be reproduced if the existing [`generic_roundtrip` test](https://github.com/kornelski/rust-security-framework/blob/ea683f291bb8dff51d3239e556c3236e09274c16/security-framework/src/passwords.rs#L203) is modified to use an empty password string. 

This bug doesn't exist on macOS, because the macOS-specific implementation doesn't retrieve data using this function.

PR #207, opened 2024-07-14
Don't extend from an empty slice.

brotskydotcom:
Fixes kornelski/rust-security-framework#206.


kornelski:
This is not the correct fix. Slices with a NULL pointer trigger UB immediately when they are created, not when they are used.

kornelski:
Workaround for https://github.com/servo/core-foundation-rs/pull/688

PR #213, opened 2024-09-17
Add CMS API support

ancwrd1:
This PR adds CMS (cryptographic message syntax) support to `security-framework-sys` and also high-level wrappers to `security-framework`.


kornelski:
Thank you

PR #217, opened 2024-11-15
SecKey: Add support for encryption and decryption

olvrlrnz:
This PR adds
- asymmetric encryption and decryption methods to `SecKey`
- searching keychains for certificates by serial number


kornelski:
Thank you

PR #218, opened 2024-12-11
Key agreement via `SecKeyCopyKeyExchangeResult()`

crashdump:
Added ` SecKeyCopyKeyExchangeResult()` to leverage `Security.framework` for ECDH-type key agreement.

### DIFF
--- a/security-framework-sys/src/item.rs
+++ b/security-framework-sys/src/item.rs
@@ -68,6 +68,11 @@ extern "C" {
     pub static kSecAttrAccessGroup: CFStringRef;
     pub static kSecAttrAccessGroupToken: CFStringRef;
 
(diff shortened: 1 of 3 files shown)


crashdump:
@kornelski, the code in this merge request works up to macOS 14, but it fails on 15+ due to the current implementation of  `SecKeyExt::from_data` relying on `SecKeyCreateFromData()` – which is deprecated.  

I have successfully implemented key derivation by using the new method in my own code, however, I stopped short of making breaking changes in this project as I wasn't sure how you'd like to handle this.

Note that the new method to import public keys supports all the platforms, so it may warrant moving `SecKeyExt` outside the `macos` mod.

Let me know your thoughts, but this should be mergeable as-is - comments welcome.

kornelski (approved):

Issue #186, opened 2023-07-08
kSecAttrAccessControl Ignored by SecItemAdd


Zack-Xb:
Hi, I have been trying to add `access control` to the function `to_dictionary` for struct `ItemAddOptions` to protect keychain entries with biometry and application password. I have also added some options in the search function, but they seem to have no effect. Note I have forked the repository and I am using a locally defined path so these changes are being used.

```
 if let Some(access_control) = &self.access_control {
            dict.add(
                &unsafe { kSecAttrAccessControl }.to_void(),
                &access_control.to_void(),
            );
        }
        ```
For example the above is where I add the option in `to_dictionary` line 599 `item.rs`. 

Has anyone gotten access control to work to protect their keychain entries?

brotskydotcom:
Hi @Zack-Xb, just thought you might want to know that it's possible to add access control to passwords using the `PasswordOptions` package in the security framework. I think that gets you what you are asking for here (except for the search part - but retrieving access controls via search is problematic for many reasons).

See also #226.

PR #220, opened 2024-12-15
Support user-defined generic password options/attributes.

### LINKED ISSUE #219: Support for using custom PasswordOptions?
Hi,

Thanks for the library! I was wondering why using `PasswordOptions` is restricted to internal use only? It has a `set_access_control_options()` which would allow us to set `kSecAccessControl` but we can't use it as `set_generic_password()` creates it's own `PasswordOptions`.

Could we add `set_generic_password_options()` and support passing a custom `PasswordOptions` when the password is created please?

I would also be keen to support setting `kSecAttrAccessible` and `kSecAttrAccessGroup` if it is ok to update `PasswordOptions` to support setters for those.

Happy to submit a PR if you are open to these changes, thanks 🙏

tmpfs:
This is a quick sketch to discuss support for custom attributes/options when creating generic passwords.

The desire is to be able to support more complex use cases such as specifying `kSecAccessControl`, `kSecAccessGroup` etc. for a generic password.

Currently I've added this as an additional `set_generic_password_options()` function so as not to introduce any breaking changes but ultimately I think a builder style API for `PasswordOptions` would be more ergonomic. Perhaps:

```rust
let options = PasswordOptions::builder()
  .service("com.example")
  .account("user-id")
  .password("mock-password")
  .access_control(AccessControlOptions::USER_PRESENCE)
  .access_group("com.example.shared-keychain-group")
  .build()
  .unwrap();  // Panic here as kSecAccessControl and kSecAccessGroup are mutually exclusive?
```

Then we could make the breaking change that `set_generic_password()` just accepts `PasswordOptions` and the password and wouldn't need the additional function if that is acceptable for the next major release.

I notice in #186 there are other people that would also like to be able to use `kSecAccessControl` with `SecItemAdd` so I think it makes sense that we support the other attributes too.

Your thoughts and feedback would be appreciated 🙏 

Closes #219

### DIFF
--- a/security-framework/src/passwords.rs
+++ b/security-framework/src/passwords.rs
@@ -27,6 +27,14 @@ pub fn set_generic_password(service: &str, account: &str, password: &[u8]) -> Re
     set_password_internal(&mut options, password)
 }
 
+/// Set a generic password using the given password options.
+/// Creates or updates a keychain entry.
+pub fn set_generic_password_options(
+    password: &[u8],
+    mut options: PasswordOptions) -> Result<()> {
+    set_password_internal(&mut options, password)
+}
+
 /// Get the generic password for the given service and account.  If no matching
 /// keychain entry exists, fails with error code `errSecItemNotFound`.
 pub fn get_generic_password(service: &str, account: &str) -> Result<Vec<u8>> {
--- a/security-framework/src/passwords_options.rs
+++ b/security-framework/src/passwords_options.rs
@@ -1,7 +1,7 @@
 //! Support for password options, to be used with the passwords module
 
 use core_foundation::{string::CFString, base::{CFType, TCFType, CFOptionFlags}, number::CFNumber};
-use security_framework_sys::{keychain::{SecProtocolType, SecAuthenticationType}, access_control::*};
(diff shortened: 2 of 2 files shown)


crashdump:
Just for info, in the key management side of things, the pattern is to use a `GenerateOptions` struct which contains all your parameters, such as access control, etc.

https://github.com/kornelski/rust-security-framework/blob/b2189b20a6f07f9dd936bb8cd263bab177bc4356/security-framework/src/key.rs#L149

```
pub struct GenerateKeyOptions {
    /// kSecAttrKeyType
    pub key_type: Option<KeyType>,
    /// kSecAttrKeySizeInBits
    pub size_in_bits: Option<u32>,
    /// kSecAttrLabel
    pub label: Option<String>,
    /// kSecAttrTokenID
    pub token: Option<Token>,
    /// Which keychain to store the key in, if any.
    pub location: Option<Location>,
    /// Access control
    pub access_control: Option<SecAccessControl>,
}
```

Validation parameters is implemented in the `GenerateKeyOptions::to_dictionary()` method.

kornelski:
Thanks

PR #221, opened 2025-01-06
Add keychain synchronisation and authentication context

crashdump:
Added the ability to pass an LAContext to the keychain key operations as well as the kSecAttrSynchronizable flag during the creation of keys.

### DIFF
--- a/security-framework/src/key.rs
+++ b/security-framework/src/key.rs
@@ -27,6 +27,8 @@ use security_framework_sys::{item::{
     kSecAttrIsPermanent, kSecAttrLabel, kSecAttrKeyType,
     kSecAttrKeySizeInBits, kSecPrivateKeyAttrs, kSecAttrAccessControl
 }};
+#[cfg(any(feature = "OSX_10_13", target_os = "ios", target_os = "tvos", target_os = "watchos", target_os = "visionos"))]
+use security_framework_sys::item::kSecAttrSynchronizable;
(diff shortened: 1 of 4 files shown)


kornelski:
Please check the build failures

crashdump:
Thanks @kornelski, I can see the `CI / Tests macos-latest tests` are also falling on the main branch, so I'll ignore that for now.

As for the `GenerateKeyOptions`,  we can either accept this as a breaking change or introduce a new builder, similar to what was suggested in MR https://github.com/kornelski/rust-security-framework/pull/220. I'll take your steer.

kornelski:
I think the cleanest solution would be to hide the new field in `GenerateKeyOptions` behind a feature flag. Although theoretically flags are supposed to be purely additive, this breakage is small enough that won't be a problem in practice.

security-framework/src/key.rs
```diff
@@ -381,6 +383,8 @@ pub struct GenerateKeyOptions {
     /// Access control
     #[deprecated(note = "use set_access_control()")]
     pub access_control: Option<SecAccessControl>,
+    /// kSecAttrSynchronizable
+    pub synchronizable: Option<bool>,
```
kornelski, line 387: the feature flag is needed only for this field.

crashdump:
I'm an idiot :facepalm: - it's fixed now.

PR #222, opened 2025-02-20
Expose simplify dict logic

zemse:
When we have a `CFDictionary`, it would be great to have a utility to cast it into HashMap for easy programmatic access. The logic is already present in this repository, this PR simply exposes it so that dependents can use it.


kornelski:
I'm trying to completely remove use of `CFDictionary` from all APIs.

I would rather not add any new methods using ObjC types in the high-level Rust crate.

zemse:
That would simplify the code on the user side.

PR #225, opened 2025-07-08
Enable access control and authentication contexts

mrdomino:
This enables creating keychain items that require Touch ID/Face ID authentication, equivalent to the Swift SecAccessControlCreateWithFlags API.

This allows things like setting BIOMETRY_ANY on created passwords, adding a public `get_generic_password_options` that allows PasswordOptions to be used for keychain items, and also allowing AccessControl and AuthenticationContext to be set on PasswordOptions.

With some arduous work (an Info.plist and entitlements plist need to be added, and the binary needs to be codesigned with a developer key for which there is a provisioning profile that allows the correct machines and features), I was able to build the included example binary and confirm that it works as expected (it shows a TouchID prompt on my Mac when I save and load the password.)

### DIFF
--- a/security-framework/Cargo.toml
+++ b/security-framework/Cargo.toml
@@ -64,6 +64,10 @@ name = "find_internet_password"
 [[example]]
 name = "set_internet_password"
 
+[[example]]
+name = "biometric_password"
+required-features = ["OSX_10_13"]
+
 [package.metadata.docs.rs]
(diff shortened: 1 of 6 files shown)


mrdomino:
I’m not bothered at all if you’d like the example removed or redone or anything; it’s LLM code, I checked that the hands all seemed to have five fingers but meh. I wanted it mainly for myself since a lot of these APIs are gated behind code-signing so I couldn’t see a way to just write unit tests for them.

Issue #231, opened 2025-08-10
Enhance ItemSearchOptions to allow skipping authenticated items


brotskydotcom:
While `ItemSearchOptions` has a way of providing an authentication context that will be reused for multiple authenticated items in a search, it lacks the basic option provided by macOS/iOS to simply skip authenticated items when doing a search. This makes it unnecessarily difficult to manage non-authenticated items, which most are.

I will submit a PR with the easy fix for this.

PR #232, opened 2025-08-10
Add ItemSearchOptions support for skipping authenticated items.

brotskydotcom:
Fixes kornelski/rust-security-framework#231.

This touches security-framework-sys (two new constants) as well as the security-framework itself.


brotskydotcom:
The macOS build failed but it doesn't look like the failure has to do with my change. All the tests run fine on my machine with stable rust and latest macOS. So I'm not sure what's up with that, sorry.

brotskydotcom:
PR #234 has been merged into this one, so once #234 is merged into the security-framework-sys project then this PR will purely affect only the security-framework project.

brotskydotcom:
Thank you!

Issue #233, opened 2025-08-15
Update item constants in security-framework-sys


brotskydotcom:
Some of the constants needed for management of secure items in macOS and iOS contexts have been around for a while but were never added to security-framework-sys.  Others are there but are marked with macOS features much higher than they need to be.

This is a request to update the following constants:

- fix version: `kSecAttrSynchronizable` (since macOS 10.9)
- add `kSecAttrSynchronizableAny`(since macOS 10.9)
- add `kSecUseAuthenticationUI` (since macOS 10.11)
- add `kSecUseAuthenticationUISkip` (since macOS 10.11)

I will submit a PR that does this.

PR #234, opened 2025-08-15
Update constants in security-framework-sys.

brotskydotcom:
Fixes kornelski/rust-security-framework#233.

This PR is only for the -sys crate. I have two outstanding PRs for the non-sys crate that I will update to match this one so they can be merged after it.


brotskydotcom:
The test failures are the same as with my other two PRs and seem to have nothing to do with these changes.

Issue #237, opened 2025-09-05
missing attributes on passwords relative to "items"


brotskydotcom:
There are two paths for adding something to secure storage in this crate:

- the `ItemAdd` and `ItemUpdate` path
- the `generic_password` and `internet_password` path

There are good reasons for these paths to be different, but it's problematic that many of the optional attributes that can be set on items cannot be set on passwords, in particular:

- `kSecAttrComment`
- `kSecAttrDescription`
- `kSecAttrLabel`
- `kSecUseDataProtectionKeychain`

Now that we have `PasswordOptions` for setting attributes on password entries, it would be good to enhance these to allow setting these attributes as well.

I will submit a PR that does this.

kornelski:
Yes, please

PR #239, opened 2025-09-07
Add more password options.

brotskydotcom:
Allows setting the desired comment, description, and label for the item.

Allows for selecting the protected data store on macOS so protected features can be used.

Fixes kornelski/rust-security-framework#237.


brotskydotcom:
Apparently this needs to be rebased. I am going to close it while I investigate.

PR #240, opened 2025-09-07
Add more password options.

brotskydotcom:
Allows setting the desired comment, description, and label for the item.

Allows for selecting the protected data store on macOS so protected features can be used.

Fixes kornelski/rust-security-framework#237.


brotskydotcom:
These changes have been tested in sandboxed-app contexts with provisioning profiles.

brotskydotcom:
So you fixed the cert tests, and I fixed the missing import. I have rebased and re-pushed to get a clean addition.

brotskydotcom:
I've seen the keychain unwrap failure before on GitHub, don't think it has anything to do with this change. Let me know if you need me to fix anything.

kornelski:
Thanks!

brotskydotcom:
Thank you (as always)! Fingers crossed that you do a crates.io release soon - I have a Mac crate that needs these changes to be released.

FYI - I am going to work on a PR to back-port the label, comment, and description attributes to the classic keychain, where they would be very useful. Also, I think it would be very useful in both legacy and protected cases to have an update-only call for passwords (which doesn't fall back to creating a new item) - that way one can update attributes without secrets (or vice versa) and not change security options (which can't be read). Once I have figured out how to do this cleanly I will submit some issues and PRs.

Issue #223, opened 2025-04-07
Implement old version fallback using `dlsym`


madsmtm:
`rustls-platform-verifier` (nowadays used in major Rust projects like Rustup) uses `security-framework`, and enables the `"OSX_10_14"` Cargo feature.

As such, trying to run `rustup update` on a machine with macOS 10.12 installed (which the Rust project otherwise still supports) results in the following dynamic linker error:

```
info: syncing channel updates for 'stable-x86_64-apple-darwin'
dyld: lazy symbol binding failed: Symbol not found: _SecTrustEvaluateWithError
  Referenced from: /Users/madsmtm/.cargo/bin/rustup
  Expected in: /System/Library/Frameworks/Security.framework/Versions/A/Security

dyld: Symbol not found: _SecTrustEvaluateWithError
  Referenced from: /Users/madsmtm/.cargo/bin/rustup
  Expected in: /System/Library/Frameworks/Security.framework/Versions/A/Security

Abort trap: 6
```

And trying to compile it manually results in a static linker error:

```
error: linking with `cc` failed: exit status: 1
  |
  = note: "cc" "$WORKSPACE/*.o" "$WORKSPACE/*.rlib" "-framework" "Security" "-liconv" "-framework" "CoreFoundation" "-lSystem" "-lc" "-lm" "-arch" "x86_64" "-mmacosx-version-min=10.12.0" "-o" "$WORKSPACE/target/debug/deps/rustls_platform_verifier-f3c84624becba9b5" "-Wl,-dead_strip" "-nodefaultlibs"
  = note: Undefined symbols for architecture x86_64:
            "_SecTrustEvaluateWithError", referenced from:
                security_framework::trust::SecTrust::evaluate_with_error::hb6ce8f9ed18e6bcb in libsecurity_framework-063e6ee960d31664.rlib(security_framework-063e6ee960d31664.security_framework.d276614c73a19b26-cgu.10.rcgu.o)
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)

error: could not compile `rustls-platform-verifier` (lib test) due to 1 previous error
```

This could of course be "fixed" in `rustls-platform-verifier` by simply not enabling the feature at all, but there's (probably) a reason why they've chosen to do so. Ideally, we should be able to use `SecTrustEvaluateWithError` when available, and only need to fall back to `SecTrustEvaluate` when not available.

[...]

PR #243, opened 2025-09-16
Make `SecTrust::evaluate_with_error` use `dlsym` when `!OSX_10_14`

madsmtm:
When `feature = "OSX_10_14"` is specified, we use the symbol `SecTrustEvaluateWithError` directly. When the feature is disabled, we first try to look it up using `dlsym`, and if the symbol doesn't exist, only then do we fall back to `SecTrustEvaluate`. This gives nicer error messages for projects that use `SecTrust::evaluate_with_error`, while still supporting macOS 10.12.

Fixes https://github.com/kornelski/rust-security-framework/issues/223 and allows fixing https://github.com/rustls/rustls-platform-verifier/issues/188 (by removing the `OSX_10_14` flag from their feature table).

We could do a similar thing at all the places where `feature = "OSX_*"` is used, however there are quite a few of those, so I'd like to only do so if there's a real need.

### DIFF
--- a/security-framework/src/trust.rs
+++ b/security-framework/src/trust.rs
@@ -8,6 +8,7 @@ use core_foundation::date::CFDate;
 use core_foundation::{declare_TCFType, impl_TCFType};
 use core_foundation_sys::base::{Boolean, CFIndex};
 
+use security_framework_sys::base::{errSecNotTrusted, errSecTrustSettingDeny};
 use security_framework_sys::trust::*;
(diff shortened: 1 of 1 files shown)


madsmtm:
Thanks!

I see that you [bumped the version](https://github.com/kornelski/rust-security-framework/commit/4d0ba3f42102a756e56c0958ddb067af4960b556), but the bump doesn't seem to have been released, was that intentional?

Issue #246, opened 2025-09-28
`SslStream` does not flush during the handshake, which causes the handshake to fail on buffered connections


opara-tor:
When `ClientBuilder` is used to establish a TLS connection on a new stream, it assumes that the underlying stream will write data immediately. But this isn't always the case:

- `Write::write()` only ensures that the data was written into the object. [`Write::flush()`](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.flush) ensures that the data has actually been written to the underlying stream.
- While TLS is typically used over unbuffered TCP connections, there are uses for TLS over other types of connections. For example [arti](https://docs.rs/arti-client/latest/arti_client/), which is a rust implementation of a Tor network client, supports forwarding any data stream (including TLS streams) over the Tor network. The Tor network sends stream data in ~510 byte chunks, so arti buffers up to this many bytes before sending them, unless a `flush()` is explicitly requested by the writer.

The `SslStream` does flush correctly when a `SslStream::flush()` is called. The problem is that it doesn't flush during the TLS handshake.

For example, I've adapted the example in the [`secure_transport`](https://docs.rs/security-framework/latest/security_framework/secure_transport/index.html#examples) module to use a buffered writer:

```rust
use security_framework::secure_transport::ClientBuilder;
use std::io::prelude::*;
use std::io::BufWriter;
use std::net::TcpStream;

fn main() {
    let stream = TcpStream::connect("google.com:443").unwrap();
    // add a buffered writer
    let stream = BufferedTcpStream::new(stream);
    let mut stream = ClientBuilder::new().handshake("google.com", stream).unwrap();

    stream.write_all(b"GET / HTTP/1.0\r\n\r\n").unwrap();
    let mut page = vec![];
    stream.read_to_end(&mut page).unwrap();
    println!("{}", String::from_utf8_lossy(&page));
}

/// Small wrapper around a `TcpStream` to provide buffered writes.
#[derive(Debug)]
struct BufferedTcpStream {
    reader: TcpStream,
    writer: BufWriter<TcpStream>,
}

impl BufferedTcpStream {
    fn new(tcp: TcpStream) -> Self {
        Self {
            writer: BufWriter::with_capacity(500, tcp.try_clone().unwrap()),
            reader: tcp,
        }
    }
}

impl Read for BufferedTcpStream {
    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
        self.reader.read(buf)
    }
}

[...]

PR #247, opened 2025-09-28
Flush the underlying stream in `write_func`

opara-tor:
This is for #246. As mentioned in that issue, I think a better solution would be to only flush automatically during the handshake. But I'm not sure the best way to do that here. Even if this solution isn't optimal, it would at least make the TLS connection work with buffered streams, where previously it would return a `Error { code: -9806, message: "connection closed via error" }`.


kornelski:
Thanks!

